PR #9, opened 2025-02-13
SUI-41 Added the logic to perform NHS Digital PDS searches

JRilforge:


shared/Shared/Models/SearchQuery.cs
```diff
@@ -0,0 +1,87 @@
...
+	[JsonPropertyName("_exact-match")]
+	public bool? ExactMatch { get; set; }
+	
+	[JsonPropertyName("_history")]
+	public bool? History { get; set; }
+	
+	[JsonPropertyName("_max-results")]
+	public bool? MaxResults { get; set; }
+
+	[RegularExpression(DateQueryRegex, ErrorMessage = "Incorrect format.")]
```
danmurray2020, line 22: Won't ever need this I don't think.

shared/Shared/Models/SearchQuery.cs
```diff
@@ -0,0 +1,87 @@
...
+	[JsonPropertyName("death-date")]
+	public string[]? DeathDate { get; set; }
+	
+	[JsonPropertyName("address-postalcode")]
+	public string? AddressPostalcode { get; set; }
+	
+	[JsonPropertyName("address-postcode")]
+	public string? AddressPostcode { get; set; }
```
danmurray2020, line 30: I think one of the above is being deprecated. Address-postcode I think.

danmurray2020:
LGTM. Just highlighted some things in the search query model, but they match the NHS API so happy to resolve those convos if we don't care so much that we would never use them.

PR #53, opened 2025-03-26
SUI-98 Removed Let's Encrypt configuration

JRilforge:
Removed configuration due to the requirement of a public domain being needed to generate a certification.

So, since the private endpoint is secure with TLS/SSL we are opting for this.

### DIFF
--- a/yarp/Yarp/Properties/launchSettings.json
+++ b/yarp/Yarp/Properties/launchSettings.json
@@ -4,7 +4,7 @@
 			"commandName": "Project",
 			"dotnetRunMessages": true,
 			"launchBrowser": false,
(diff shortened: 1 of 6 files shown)


yarp/Yarp/Properties/launchSettings.json
```diff
@@ -4,7 +4,7 @@
 			"commandName": "Project",
 			"dotnetRunMessages": true,
 			"launchBrowser": false,
-			"applicationUrl": "http://localhost:5000;https://*:5443",
+			"applicationUrl": "http://localhost:5100",
```
danmurray2020, line 7: Do we only want to support https now though?

danmurray2020:
Do we want to enforce that it will only work using https in this PR?

PR #68, opened 2025-04-17
Feat: 146 client watcher configurations

StuwiiDev:
- Updated PDS Client watcher to have a URL input arg. This is because each LA will have their own unique matching service URL. Also updated to have some build details for when it is packaged as NuGet.
- Updated DBS Client watcher to have some build details for when it is packaged as NuGet.
- Added README for developers
- Added a client user README for instructions on how to use this.

Out of scope on this PR
- Nuget feed from pipeline. This will be done in a following PR once we get the OK from DfE to publish to Nuget.org

### DIFF
--- a/README_Clients_User_Manual.md
+++ b/README_Clients_User_Manual.md
@@ -1,0 +1,51 @@
+# Readme for downloading and running the SUI Client Watcher tools from NuGet source
+
+## Prerequisites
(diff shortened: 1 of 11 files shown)


StuwiiDev:
Just wondering whether we should control version through hard coding in the csproj or purely through CI

PR #71, opened 2025-04-24
Feat: 144 - data pds differences

StuwiiDev:
- Updated the Get and Set Activity Tag to be 'Baggage' instead as this carries data accross HTTP requests via headers. This change was made so we can have the SearchId persist to the External service.
- Added a Comparison class and unit tested some scenarios
- Added logging to the FhirService as this is the closest place to the request data.

I believe I've captured all required fields but please double check when reviewing this PR.

### DIFF
--- a/SUI.Core/Endpoints/NhsFhirClient.cs
+++ b/SUI.Core/Endpoints/NhsFhirClient.cs
@@ -63,19 +63,14 @@ public class NhsFhirClient(ITokenService tokenService,
             }
             else
             {
+                logger.LogInformation("{EntryCount} patient record(s) found", patient.Entry.Count);
                 if (patient.Entry.Count == 0)
                 {
-                    logger.LogInformation("No patient record found");
-
                     return SearchResult.Unmatched();
                 }
 
                 if (patient.Entry.Count == 1)
                 {
-                    var birthDate = patient.Entry[0].Resource["birthDate"].ToString();
-                    var gender = patient.Entry[0].Resource["gender"].ToString();
-
-                    logger.LogInformation("1 patient record found: BirthDate={BirthDate} Gender={Gender}", birthDate, gender);
                     LogInputAndPdsDifferences(query, (Patient)patient.Entry[0].Resource);
 
                     return SearchResult.Match(patient.Entry[0].Resource.Id, patient.Entry[0].Search.Score);
--- a/SUI.Core/Endpoints/NhsFhirClient.cs
+++ b/SUI.Core/Endpoints/NhsFhirClient.cs
@@ -6,6 +6,7 @@ using Shared.Models;
 using SUI.Core.Endpoints.AuthToken;
 using System.Collections;
 using System.Net.Http.Headers;
+using SUI.Core.Services;
 
 namespace SUI.Core.Endpoints;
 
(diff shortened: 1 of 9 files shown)


SUI.Core/Endpoints/NhsFhirClient.cs
```diff
@@ -68,17 +69,19 @@ public class NhsFhirClient(ITokenService tokenService,
...
+                if (patient.Entry.Count == 1)
                 {
-                    var birthDate = patient.Entry[0].Resource["birthDate"];
-                    var gender = patient.Entry[0].Resource["gender"];
-                
-                    logger.LogInformation($"1 patient record found: BirthDate={birthDate} Gender={gender}");
+                    var birthDate = patient.Entry[0].Resource["birthDate"].ToString();
+                    var gender = patient.Entry[0].Resource["gender"].ToString();
+
+                    logger.LogInformation("1 patient record found: BirthDate={BirthDate} Gender={Gender}", birthDate, gender);
```
danmurray2020, line 78: Unsure if we should log the birthdate. Potential to link back to a real person if you have the origianal dataset.

SUI.Core/Services/FieldComparerService.cs
```diff
@@ -0,0 +1,44 @@
...
+    public static List<string> ComparePatientFields(SearchQuery query, Patient patient)
+    {
+        var differentFields = new List<string>();
+
+        CompareField(query.Birthdate?.FirstOrDefault(), $"eq{patient.BirthDate}", nameof(SearchQuery.Birthdate), differentFields);
+        CompareField(query.AddressPostalcode, patient.Address.FirstOrDefault()?.PostalCode, nameof(SearchQuery.AddressPostalcode), differentFields);
+        CompareField(query.Gender, patient.Gender?.ToString(), nameof(SearchQuery.Gender), differentFields);
+        CompareField(query.Family, patient.Name.FirstOrDefault()?.Family, nameof(SearchQuery.Family), differentFields);
+        CompareField(query.Given?.FirstOrDefault(), patient.Name.FirstOrDefault()?.Given.FirstOrDefault(), nameof(SearchQuery.Given), differentFields);
+        CompareField(query.GeneralPractitioner, patient.GeneralPractitioner.FirstOrDefault()?.Identifier.Value, nameof(SearchQuery.Given), differentFields);
```
danmurray2020, line 23: We don't ever get given this as input date.

SUI.Core/Endpoints/NhsFhirClient.cs
StuwiiDev: That existed already, I agree, it should probably be removed

SUI.Core/Services/FieldComparerService.cs
StuwiiDev: o the GP? it's in the searchQuery but I can remove

StuwiiDev: Updated: remove the search field from here and the SearchQuery.cs to avoid future confusion. YAGNI

SUI.Core/Endpoints/NhsFhirClient.cs
StuwiiDev: I've updated to not log the values. Only the number of records found

danmurray2020 (approved):

PR #72, opened 2025-04-24
Chore: 175 - resolve build warnings

StuwiiDev:
Removed all build warnings.
Updated some syntax

There were some AppHost SDK inclusions in non AppHost applications as well which were causing warnings.


danmurray2020 (approved):
LGTM

PR #73, opened 2025-04-28
Feat: 132 - infra parametarize

StuwiiDev:
- Moved DfE specific variable values to params in the bicep files
- Updated github CI so pass our variables in that were originally in the bicep files
- Added a what-if. This is 1 step short of a approval gate but want to see if this works for now

When checking this PR. Please be thorough in ensuring I have caught all DfE specific vars. I am ensure about logging/resource specific names that come after the prefix and environment name (e.g. `{prefix}-{environmentname}-01`)

### DIFF
--- a/.github/workflows/gh-deploy-infra.yml
+++ b/.github/workflows/gh-deploy-infra.yml
@@ -29,6 +29,7 @@ jobs:
       AZURE_LOCATION: "westeurope"
       AZURE_RESOURCE_GROUP: ${{ secrets.AZURE_RESOURCE_GROUP }}
       APPLICATION_VERSION: ${{ inputs.version }}
+      AZURE_LOG_ANALYTICS_WORKSPACE_ID: '/subscriptions/8fc7ea96-7305-492f-85f7-09069bb8fd29/resourceGroups/s215d01-integration/providers/Microsoft.OperationalInsights/workspaces/s215d01-integration-loganalytics-01'
 
     steps:
       - name: Checkout
--- a/.github/workflows/gh-deploy-infra.yml
+++ b/.github/workflows/gh-deploy-infra.yml
@@ -9,6 +9,10 @@ on:
         type: choice
         options:
           - 'Integration'
+      what_if:
+        description: Run bicep what-if to check for changes
+        default: true
+        type: boolean
       version:
         description: The version of the infrastructure to deploy
         default: 'v0.0.1'
@@ -55,6 +59,12 @@ jobs:
 
       - name: Install .NET Aspire workload
         run: dotnet workload install aspire
+        
+      - name: Ensure Azure CLI is installed
+        run: |
+          if ! command -v az &> /dev/null; then
+            curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
+          fi
 
       - name: Log in with Azure (Federated Credentials)
         if: ${{ env.AZURE_CLIENT_ID != '' }}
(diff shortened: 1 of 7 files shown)


.github/workflows/gh-deploy-infra.yml
```diff
@@ -23,12 +29,16 @@ jobs:
...
       AZURE_ENV_NAME: ${{ inputs.environment }}
+      AZURE_ENV_PREFIX: ${{ inputs.environnment_prefix }}
       AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
       AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
       AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
-      AZURE_LOCATION: "westeurope"
+      AZURE_LOCATION: 'westeurope'
       AZURE_RESOURCE_GROUP: ${{ secrets.AZURE_RESOURCE_GROUP }}
       APPLICATION_VERSION: ${{ inputs.version }}
+      AZURE_LOG_ANALYTICS_WORKSPACE_ID: '/subscriptions/8fc7ea96-7305-492f-85f7-09069bb8fd29/resourceGroups/s215d01-integration/providers/Microsoft.OperationalInsights/workspaces/s215d01-integration-loganalytics-01'
```
danmurray2020, line 39: I think you can reference that as an output of the resource rather than having to supply it.

app-host/infra/secrets/secrets.module.bicep
```diff
@@ -1,9 +1,11 @@
 @description('The location for the resource(s) to be deployed.')
 param location string = resourceGroup().location
 
+param environmentPrefix string
+
 // 3 - 24 alphanumeric characters 
 resource secrets 'Microsoft.KeyVault/vaults@2023-07-01' = {
-  name: 's215d01-int-kv01'
+  name: '${environmentPrefix}-int-kv01'
```
danmurray2020, line 8: That int is for integration, just needs the env in there.

danmurray2020:
I think we would want to also pass through the networks. So the values: 
@description('The address prefix for the virtual network')
param containerAppVnet string = '192.168.0.0/25'

@description('Container App environment subnet')
param containerAppEnvSubnet string = '192.168.0.0/26'

@description('Container App environment subnet')
param containerAppFirewallSubnet string = '192.168.0.64/26'

danmurray2020:
The values after the prefixes are just meant to be identifying names that match the resource. Should be fine to leave. Unfortunately Azure has different naming standards for different resources (no underscores, 10 chars, etc.) hence why some don't seem to follow a standard.

.github/workflows/gh-deploy-infra.yml
StuwiiDev: o it's already got the Id in the resources from creating the analytics, so need to supply it manually. this came about from a update on top of @JRilforge last PR where it was manually added

app-host/infra/secrets/secrets.module.bicep
StuwiiDev: Added lowercase and substring to replace the 'int' with environment name

StuwiiDev:
> I think we would want to also pass through the networks. So the values: @description('The address prefix for the virtual network') param containerAppVnet string = '192.168.0.0/25'
> 
> @description('Container App environment subnet') param containerAppEnvSubnet string = '192.168.0.0/26'
> 
> @description('Container App environment subnet') param containerAppFirewallSubnet string = '192.168.0.64/26'

As discussed, not actionin this comment in this PR.

danmurray2020 (approved):

PR #91, opened 2025-05-06
SUI-36 migrated non test code to `src` folder

JRilforge:


SUI.Core/SUI.Core.csproj
```diff
@@ -18,8 +18,8 @@
 	</ItemGroup>
 
 	<ItemGroup>
-	  <ProjectReference Include="..\shared\Shared\Shared.csproj" />
-	  <ProjectReference Include="..\SUI.Types\SUI.Types.csproj" />
+	  <ProjectReference Include="..\src\Shared\Shared.csproj" />
+	  <ProjectReference Include="..\src\Shared\Shared.csproj" />
```
danmurray2020, line 22: same reference twice?

SUI.Core/Services/MatchingService.cs
```diff
@@ -7,10 +7,10 @@
 using Newtonsoft.Json;
 
 using Shared.Models;
+using Shared.Models.Client;
```
danmurray2020, line 10: seems strange that the matching service has a dependency on something from the client.

SUI.Core/SUI.Core.csproj
```diff
@@ -18,8 +18,8 @@
 	</ItemGroup>
 
 	<ItemGroup>
-	  <ProjectReference Include="..\shared\Shared\Shared.csproj" />
-	  <ProjectReference Include="..\SUI.Types\SUI.Types.csproj" />
+	  <ProjectReference Include="..\src\Shared\Shared.csproj" />
```
danmurray2020, line 21: I thought we would be getting rid of SUI.Core?

SUI.Core/Services/MatchingService.cs
JRilforge: changed

SUI.Core/SUI.Core.csproj
JRilforge: this is just the first PR, as you lot mentioned in the chat, it would be too complex to do it all in one PR

danmurray2020 (approved):

StuwiiDev:
minor comment. - could the 'sui-tests' folder be changed to 'tests' I can't see a need to add sui onto it, 'tests' is the standard convention

JRilforge:
> minor comment. - could the 'sui-tests' folder be changed to 'tests' I can't see a need to add sui onto it, 'tests' is the standard convention. It can be a seperate PR if that's easier

@danmurray2020 That's the plan for the next PR

PR #112, opened 2025-05-14
SUI-36 Migrated `SUI.Core` members to Shared or to their related app projects

JRilforge:
- Changed `DBS.Client` to `DBS.Response.Logger`


src/app-host/infra/resources.bicep
```diff
@@ -545,69 +545,199 @@ resource applicationInsights 'Microsoft.Insights/components@2020-02-02' = {
...
+        }
+      ]
+    }
+    retentionInDays: 30
+  }
+}
+
 param dataCollectionRules_DbsClientConsoleAppLogsRule_name string = 'DbsClientConsoleAppLogsRule'
 
-// resource dataCollectionRules_DbsClientConsoleAppLogsRule_name_resource 'Microsoft.Insights/dataCollectionRules@2023-03-11' = {
```
danmurray2020, line 550: It might be a good time just to get rid of all the commented out resources. I know I probably put them in... Apologies.

danmurray2020 (commented):
Just wondering whether SUI.DBS.Response.Logger needs to be three separate projects or could be moved into one project? Unsure whether we think the other parts of it would be useful else where.

JRilforge:
> Just wondering whether SUI.DBS.Response.Logger needs to be three separate projects or could be moved into one project? Unsure whether we think the other parts of it would be useful else where.

@danmurray2020 Can watcher and console be in one project? watcher is still necessary and console is used primarily for smoke testing (to my knowledge)

danmurray2020 (approved):

PR #118, opened 2025-05-14
SUI-4 Added a json converter to convert MatchStatus and QualityType into human readable strings instead of ints

JRilforge:
a local test for the desired outcome:

Screenshot 2025-05-14 at 14 39 46


danmurray2020:
looks like a lot of the smoke test records are deleted?

StuwiiDev:
why didnt you use [JsonConverter(typeof(JsonStringEnumConverter))]?

JRilforge:
> looks like a lot of the smoke test records are deleted?

@danmurray2020 that csv i changes is just a duplicate. the smoke test csv can be found in dir `tests/E2E.Tests/Resources/Smoke`

StuwiiDev:
> > looks like a lot of the smoke test records are deleted?
> 
> @danmurray2020 that csv i changes is just a duplicate. the smoke test csv can be found in dir `tests/E2E.Tests/Resources/Smoke`

You created it globally, but only used it in for  PersonMatchResponse, I was curious why you didn't apply it to the enum properties instead, save needing to add the options to every convert

JRilforge:
> > > looks like a lot of the smoke test records are deleted?
> > 
> > 
> > @danmurray2020 that csv i changes is just a duplicate. the smoke test csv can be found in dir `tests/E2E.Tests/Resources/Smoke`
> 
> You created it globally, but only used it for PersonMatchResponse, I was curious why you didn't apply it to the enum properties instead, save needing to add the options to every convert

@StuwiiDev this approach is future proofed in case of changes or future additional endpoints. and it's less code.

danmurray2020:
> > looks like a lot of the smoke test records are deleted?
> 
> @danmurray2020 that csv i changes is just a duplicate. the smoke test csv can be found in dir `tests/E2E.Tests/Resources/Smoke`

Do we need a duplicate csv? Not sure I understand why that file is there then.

JRilforge:
> > > looks like a lot of the smoke test records are deleted?
> > 
> > 
> > @danmurray2020 that csv i changes is just a duplicate. the smoke test csv can be found in dir `tests/E2E.Tests/Resources/Smoke`
> 
> Do we need a duplicate csv? Not sure I understand why that file is there then.

I used it here when manually testing:
image
It's better than the original `test_success.csv`. I've transferred this file contents to `test_success.csv` now

StuwiiDev (approved):

PR #129, opened 2025-05-22
SUI-43 Created a E2E Test to run against the actual NHS PDS api

JRilforge:


StuwiiDev:
Does PDS/NHS have any policies about using automated test tools against their API's? wondering if it's ok to do so

PR #137, opened 2025-05-28
SUI-407 Switch smoke tests use of console app to watcher app

JRilforge:


.github/workflows/gh-client-build.yml
```diff
@@ -96,20 +86,29 @@ jobs:
...
-              $blobName = "suiw.exe"
-              $destinationDir = "C:\Users\AzCopy\${{ env.BLOB_STORAGE_LOCATION }}"
-              $destinationPath = "$destinationDir\$blobName"
+              $fileName = "suiw.exe"
+              $destinationDir = "C:\Users\SmokeTests\"
+              $destinationPath = "$destinationDir\$fileName"
             
               # Create Directory if does not exist
-              New-Item -ItemType Directory -Path "C:\Users\AzCopy\${{ env.BLOB_STORAGE_LOCATION }}"
+              New-Item -ItemType Directory -Path "C:\Users\SmokeTests\"
```
StuwiiDev, line 94: use $destinationDir here

danmurray2020:
I was hoping to get rid of .github/workflows/gh-dbs-response-logger-build.yml and .github/workflows/gh-client-build.yml as they have been superseded by https://github.com/DFE-Digital/SUI_Matcher/blob/main/.github/workflows/gh-publish-nuget.yml. 

I also think it would be a good idea to setup and rip down the smoke tests every time they run. I'll incorporate some of these changes into my PR and see what you think.

PR #140, opened 2025-05-29
SUI-426 Added Query Algorithm Versioning

JRilforge:
Opted for manual version tracking.

- The version is stored in a static variable.
- The static variable persisted through the set of aspire apps.
- Then logged the output.

----

- ~Using .csx file (c# script) to check/manage the algorithm versioning, then leveraging the git pre-commit hook to execute it~

image

- ~Evidence of algorithm version displaying in the logs:~

Screenshot 2025-05-30 at 13 07 42


src/matching-api/Util/AlgorithmVersionUtil.cs
```diff
@@ -0,0 +1,92 @@
...
+        string prevHash = versionJson?.PreviousHash ?? string.Empty;
+        int version = versionJson?.Version ?? 1;
+
+        if (newHash == currentHash) // keep version the same
+        {
+            Console.WriteLine($"No changes detected, version remains {version}");
+        }
+        else
+        {
+            if (newHash == prevHash) // revert to previous version
```
danmurray2020, line 44: I guess this works if we revert to a version one before, but we wouldn't be aware if we moved to two versions before. I guess we might not need that unless we were going to keep all the hashes?

tests/Unit.Tests/Matching/QueryAlgorithmBuildTests.cs
```diff
@@ -0,0 +1,93 @@
+﻿using MatchingApi.Util;
+
+using Newtonsoft.Json;
+
+using AlgoVersion = MatchingApi.Util.AlgorithmVersionUtil.AlgoVersion;
+
+namespace Unit.Tests.Matching;
+
+public sealed class QueryAlgorithmBuildTests
```
StuwiiDev, line 9: You need a test for it creating a new file if one does not exist (first time run)

tests/Unit.Tests/Matching/QueryAlgorithmBuildTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        // Act
+        AlgorithmVersionUtil.StoreOrIncrementAlgorithmVersion();
+
+        // Assert
+        var latestVersionJson = JsonConvert.DeserializeObject<AlgoVersion>(File.ReadAllText(versionFile));
+        Assert.Equal(1, latestVersionJson?.Version);
+    }
+
+    [Fact]
+    public void CurrentAlgorithmVersionReturned()
```
StuwiiDev, line 75: naming nitpick - could 'GetCurrentAlgorithmVersion' be in the name, so it's clear it's a different method test to the others?

src/matching-api/Util/AlgorithmVersionUtil.cs
JRilforge: opted for the manual approach, so removed

tests/Unit.Tests/Matching/QueryAlgorithmBuildTests.cs
JRilforge: opted for the manual approach, so removed

StuwiiDev:
In the ticket you have removed the part about rolling back versions. What has changed that decision?

tests/Unit.Tests/Matching/MatchingServiceTests.cs
```diff
@@ -280,6 +293,52 @@ public async Task ShouldLogPersonSpecificationAndResultStatus_WithMatchCompleted
...
+        await subj.SearchAsync(model);
+
+        // Assert
+        var logMessages = writer.ToString().Split("\n");
+
+        Assert.NotEmpty(logMessages);
+
+        outputHelper.WriteLine(writer.ToString());
+
+        Assert.True(logMessages.Count(x => x.Contains($"[Algorithm=v{MatchingService.AlgorithmVersion}]")) >= 1);
```
StuwiiDev, line 340: minor - sonar warning here to use any instead of count, performance

PR #146, opened 2025-06-03
Bugfix/no ref/wrong file

danmurray2020:
Moving the steps to auto deploy from client to application action.

### DIFF
--- a/.github/workflows/gh-deploy-infra.yml
+++ b/.github/workflows/gh-deploy-infra.yml
@@ -53,10 +53,12 @@ jobs:
         id: get_latest_tag
         run: |
           if [[ -z "${{ env.APPLICATION_VERSION }}" ]]; then
+            git fetch --tags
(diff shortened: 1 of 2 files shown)


.github/workflows/gh-deploy-infra.yml
```diff
@@ -41,15 +46,24 @@ jobs:
...
-        if: ${{ github.event.inputs.version != '' }}
-        with:
-          ref: ${{ env.APPLICATION_VERSION }}
 
-      - name: Checkout (branch)
+      - name: 'Get latest tag if no version provided'
+        id: get_latest_tag
+        run: |
+          if [[ -z "${{ env.APPLICATION_VERSION }}" ]]; then
+            LATEST_TAG=$(git describe --tags `git rev-list --tags --max-count=1`)
```
JRilforge, line 56: there needs to be a `git fetch --tags` above this line, otherwise you'll get `fatal: No names found, cannot describe anything.`

PR #187, opened 2025-07-14
Poc - SUI-565 - audit logger

StuwiiDev:
This is a proof of concept to add a AuditLogger to the application. This is by no means a final version and will be updated as we gather more requirements.

Technologies added:
- Azure storage tables
- Feature flag management service

This feature is behind a feature flag in both Infrastructure and Application code so that we can merge into main without effecting existing workflow. Mosty so we don't have a long life branch that could go stale and/or be difficult to rebase.

To test locally:
- Change the launch settings feature flag to 'true' and then run the app. It should work first time. It will create a azurite docker image for storage. Then you can run the matching API and see if logs appear in the Audit table.

TODO:
- [x] Test in CI and azure

### DIFF
--- a/.github/workflows/gh-deploy-app.yml
+++ b/.github/workflows/gh-deploy-app.yml
@@ -20,6 +20,10 @@ on:
         description: The prefix for the environment
         default: 's215d01'
         type: string
+      enable_audit_logging:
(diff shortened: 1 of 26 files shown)


tomhawkin (dismissed):
looks good

tomhawkin (approved):

PR #205, opened 2025-07-31
SUI-630 Updated the algorithm to return the best query result

JRilforge:
- Now performs several of the queries, we stop on the desired result (status: Match, confidence score: 1), otherwise we carry on and return the best of the bunch.

- To improve the readability of the process stages in the logs, query codes were added instead of the index numbers.

### DIFF
--- a/src/matching-api/Services/MatchingService.cs
+++ b/src/matching-api/Services/MatchingService.cs
@@ -16,7 +16,7 @@ public class MatchingService(
     IValidationService validationService,
     IAuditLogger auditLogger) : IMatchingService
 {
-    public static readonly int AlgorithmVersion = 2;
+    public static readonly int AlgorithmVersion = 3;
 
     public async Task<PersonMatchResponse> SearchAsync(PersonSpecification personSpecification)
     {
@@ -33,9 +33,6 @@ public class MatchingService(
         var validationResults = validationService.Validate(personSpecification);
 
         var dataQualityResult = DataQualityEvaluatorService.ToQualityResult(personSpecification, validationResults.Results!.ToList());
-
-
-
 
         if (!HasMinDataRequirements(dataQualityResult))
         {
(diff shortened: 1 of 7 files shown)


src/matching-api/Services/MatchingService.cs
```diff
@@ -180,47 +177,57 @@ private static SearchQuery[] GetSearchQueries(PersonSpecification model)
...
-        var queries = new List<SearchQuery>
+        var queryOrderedMap = new OrderedDictionary<string, SearchQuery>
         {
-            new() // exact search on only given, family and dob
             {
-                ExactMatch = true,
-                Given = modelName,
-                Family = model.Family,
-                Birthdate = dob
+                "ExactGFD", new() // exact search on only given, family and dob
```
danmurray2020, line 183: has the spacing increased to 8 here?

src/matching-api/Services/MatchingService.cs
```diff
@@ -287,22 +300,51 @@ private async Task<MatchResult2> MatchAsync(PersonSpecification model)
                     {
                         status = MatchStatus.PotentialMatch;
                     }
+                    else
+                    {
+                        status = MatchStatus.NoMatch;
```
danmurray2020, line 305: I think status will be noMatch if not changed, so may not need to do this?

src/matching-api/Services/MatchingService.cs
```diff
@@ -287,22 +300,51 @@ private async Task<MatchResult2> MatchAsync(PersonSpecification model)
...
+        {
+            logger.LogInformation("Search query ({Query}) resulted in status '{Status}'", currentQueryCode, currentStatus);
+
+            return new MatchResult2(currentSearchResult, currentStatus, currentQueryCode);
+        }
+
+        logger.LogInformation("Search query ({QueryLength}) resulted in status 'NoMatch'", queries.Count - 1);
 
-        return new MatchResult2(MatchStatus.NoMatch, queries.Length - 1);
+        return new MatchResult2(MatchStatus.NoMatch, (queries.Count - 1).ToString());
```
danmurray2020, line 348: should this now be querycode?

danmurray2020:
Think sonarcloud has spotted two issues which should be relatively easy to solve to I think.

JRilforge:
> Think sonarcloud has spotted two issues which should be relatively easy to solve to I think.

Resolved

PR #239, opened 2025-10-02
updated to the latest fhir client

tomhawkin:


src/external-api/Services/NhsFhirClient.cs
```diff
@@ -128,8 +124,13 @@ private static DemographicResult DemographicResult(string fhirError = "", Status
         };
     }
 
-    private void LogInputAndPdsDifferences(SearchQuery query, Patient patient)
+    private void LogInputAndPdsDifferences(SearchQuery query, Patient? patient)
     {
+        if (patient == null)
```
StuwiiDev, line 129: Is this used in any branches where patient is null? might be better to keep a null check outside of this method so it's simpler

tests/Unit.Tests/External/NhsFhirClientTests/PerformSearchTests.cs
```diff
@@ -54,6 +54,26 @@ public async Task ShouldGetSearchResultsMultiMatched_WhenMultipleMatches()
         Assert.Equal(SearchResult.ResultType.MultiMatched, result.Type);
     }
 
+    [Fact]
+    public async Task ShouldGetSearchResultsError_WhenMatchTypeNotMultipleMatches()
```
StuwiiDev, line 58: I'm not sure the title of this test makes sense, implies if it's not a multimatch it will be an error? 
Theres only 2 ways to get an error, exception from Fhir or we get more than 1 patient (which I don't think is possible anyway in application mode)

tomhawkin: i have changed it to something better

PR #240, opened 2025-10-07
Feature/s UI 788

tomhawkin:


src/Shared/Models/DemographicsRequest.cs
```diff
@@ -0,0 +1,12 @@
...
+
+using Shared.Attributes;
+
+namespace Shared.Models;
+
+public sealed class DemographicsRequest
+{
+    [Required]
+    [JsonPropertyName("nhs-number")]
+    public required string NhsNumber { get; set; }
```
Ruagit, line 11: Do we need validation here like we have in the DemographicRequest.cs file for the nhs number?

tests/Unit.Tests/Matching/ReconciliationServiceTests.cs
StuwiiDev, line 1: could the test names be more explicit in the fact they are asserting on field values existing or missing as appose to differences. This should make it much clearer what the output of the Reconcile service is bringing back in it's DifferencesString. 

The test name now makes sense to the property name but not it's purpose

tomhawkin: i didn't realise you wanted all the tests in that file renaming

PR #301, opened 2026-03-31
feat: job processor v1 fa

StuwiiDev:
Skeleton first PR - not all functionality added. Just enough to get it to run.

- Added azure function app
- Linked it to Aspire for local dev
- Reads from a storage queue and blob
- Added a test script for ease of local testing
- Added some very basic unit tests prior to more in depth functionality


src/SUI.Client.StorageProcessFunction/SUI.Client.StorageProcessFunction.csproj
```diff
@@ -0,0 +1,28 @@
...
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage.Queues" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" OutputItemType="Analyzer" />
+    <PackageReference Include="Azure.Storage.Blobs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Shared\Shared.csproj" />
+    <ProjectReference Include="..\SUI.Client\SUI.Client.Core\SUI.Client.Core.csproj" />
```
jtaylor100: I think if the SUI.Client.Core project is going to be referenced, then this project needs to sit in the `SUI.Client` folder. Or otherwise, the projects in that folder need taking to the top-level. If the folders there, it's worth making it mean something (All of the 'client' code, collected) or to get rid of it.

Can take into another PR if it screws any of your changes since creating this PR.

src/SUI.Client/SUI.Client.StorageProcessFunction/Application/BlobPayloadProcessor.cs
```diff
@@ -0,0 +1,23 @@
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Extensions.Logging;
+
+namespace SUI.StorageProcessFunction.Application;
+
+[ExcludeFromCodeCoverage(Justification = "Function not implemented yet")]
+public sealed class BlobPayloadProcessor(ILogger<BlobPayloadProcessor> logger)
```
jtaylor100, line 7: Hmm, I'd consider everything in the 'Application' namespace here to be more a part of the 'Infrastructure' layer.

Where the 'Infrastructure' layer stops and the 'Application' layer starts should be where you have the data to be processed for a particular use case in a format which is common and reused over different sets of infrastructure. Which in this case would be a table of input data and maybe some metadata.

Once you have that data in a common format, you can execute the use case itself which is to start querying the matching service for results.

What do you think? If you've got any definitions for the layers you're working with, it could help me understand. Maybe we should ADR this.

src/app-host/Program.cs
```diff
@@ -76,4 +79,23 @@
...
+    builder
+        .AddProject<Projects.SUI_Client_StorageProcessFunction>("storage-process-function")
+        .WithEnvironment(ctx =>
+        {
+            ctx.EnvironmentVariables["AzureWebJobsStorage"] = storageConnectionString;
+            ctx.EnvironmentVariables["FUNCTIONS_WORKER_RUNTIME"] = "dotnet-isolated";
+            ctx.EnvironmentVariables["QueueName"] = "storage-process-job";
+        });
+}
+
```
IgiCodes, line 122: **issue (blocking):** I think this breaks the default local-dev path. `EnableStorageProcessFunction` is `true` in development, and we set `UseDevelopmentStorage=true` for the function here, but the AppHost only provisions/runs Azure Storage inside the `EnableAuditLogging` branch above (which is false by default). That means Aspire can start the function without any Azurite/storage resource behind it I think? Maybe the Aspire setup handles this in a way I'm not familiar with.

Could we either provision storage independently for this feature, or avoid enabling the function by default until that wiring exists?

src/SUI.Client.StorageProcessFunction/Application/StorageQueueMessageProcessor.cs
```diff
@@ -0,0 +1,69 @@
...
+        if (string.IsNullOrWhiteSpace(fileName))
+        {
+            throw new InvalidOperationException(
+                "Queue message blobName did not contain a file name."
+            );
+        }
+
+        var timestamp = timeProvider
+            .GetUtcNow()
+            .ToString("yyyyMMddHHss", CultureInfo.InvariantCulture);
```
IgiCodes: **issue (blocking):** The timestamp format here is missing minutes: `yyyyMMddHHss`. That means two runs for the same filename at `12:01:05` and `12:59:05` will generate the same archive path.

Should this be `yyyyMMddHHmmss` (or otherwise made unique)?

tests/Unit.Tests/StorageProcessFunction/StorageQueueMessageProcessorTests.cs
```diff
@@ -0,0 +1,128 @@
...
+            .ReturnsAsync(blobFile);
+
+        await _sut.ProcessAsync(queueMessage, CancellationToken.None);
+
+        _blobFileReader.Verify(x => x.ReadAsync(queueMessage, CancellationToken.None), Times.Once);
+        _blobPayloadProcessor.Verify(
+            x => x.ProcessAsync(blobFile, CancellationToken.None),
+            Times.Once
+        );
+        var utcNow = $"{_timeProvider.GetUtcNow():yyyyMMddHHss}_test-file/test-file.csv";
```
IgiCodes: **issue (blocking):** This test is currently using the same timestamp format that is missing the minutes. It will need updating alongside the implementation 🙂

IgiCodes (dismissed):
Looks great so far!
But I think there are two blocking issues before merge: the new local-dev Aspire path doesn't appear to provision storage for the function, and the processed blob naming is missing minutes from the timestamps.

src/app-host/Program.cs
IgiCodes: Sorry, this comment was meant to be on the entire block, not just line 100 🫠

src/SUI.Client/SUI.Client.StorageProcessFunction/host.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "2.0",
+  "functionTimeout": "02:00:00",
+  "extensions": {
+    "queues": {
+      "visibilityTimeout": "00:01:00",
+      "maxDequeueCount": 5,
+      "messageEncoding": "none"
+    }
```
jtaylor100: Thoughts? Feels like parallel processing could introduce concurrency headaches that we don't need to account for, because it's a batch system and performance isn't critical.

Alternatively we could set `newBatchThreshold` to 1 because I don't think we're ever going to get more than 3 queue items simultaneously. By the default configuration [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-queue?tabs=isolated-process%2Cextensionv5&pivots=programming-language-csharp#host-json) it looks like the function would only start processing messages when the queue reaches 8 items. Idk you might have gotten around this somewhere else.

```suggestion
      "messageEncoding": "none",
      "batchSize": 1
    }
```

src/SUI.Client/SUI.Client.StorageProcessFunction/Functions/ProcessStorageQueueMessageFunction.cs
```diff
@@ -0,0 +1,22 @@
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+using SUI.StorageProcessFunction.Application;
+
+namespace SUI.StorageProcessFunction.Functions;
+
+public sealed class ProcessStorageQueueMessageFunction(
+    IStorageQueueMessageProcessor processor,
+    ILogger<ProcessStorageQueueMessageFunction> logger
+)
```
jtaylor100, line 11: Josh's 20 questions: How does this get picked up on startup as the entry point for the function? I couldn't see it referenced in `Program.cs`. Guess it uses some reflection magic?

src/SUI.Client/SUI.Client.StorageProcessFunction/Application/StorageQueueMessageProcessor.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace SUI.StorageProcessFunction.Application;
+
+public sealed class StorageQueueMessageProcessor(
+    ILogger<StorageQueueMessageProcessor> logger,
+    TimeProvider timeProvider,
+    IBlobFileReader blobFileReader,
+    IBlobPayloadProcessor blobPayloadProcessor
+) : IStorageQueueMessageProcessor
+{
+    private const string ProcessedContainerName = "processed";
```
jtaylor100: Nit: Might be worth adding this to the configuration, so it can be passed in through IaC. Though you and Brad can work out if that's a good idea.

src/SUI.Client/SUI.Client.StorageProcessFunction/Infrastructure/AzureStorage/AzureBlobFileReader.cs
```diff
@@ -0,0 +1,62 @@
...
+    public async Task<BlobFileContent> ReadAsync(
+        StorageBlobMessage blobMessage,
+        CancellationToken cancellationToken
+    )
+    {
+        var blobClient = blobServiceClient
+            .GetBlobContainerClient(blobMessage.ContainerName!)
+            .GetBlobClient(blobMessage.BlobName!);
+
+        var download = await blobClient.DownloadContentAsync(cancellationToken);
```
jtaylor100, line 22: I guess there's no risk attached to downloading it into memory? Bearing in mind, in our new LA we might have files 2.5x larger than our existing one, due to a larger sample size.

src/SUI.Client/SUI.Client.StorageProcessFunction/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Queue Message Contract
+
+The queue message body must be JSON with this shape:
+
+```json
+{
+  "containerName": "incoming",
+  "blobName": "test-file.csv"
+}
```
jtaylor100: Guessing this is what Event Grid generates?

tests/Unit.Tests/StorageProcessFunction/StorageQueueMessageProcessorTests.cs
```diff
@@ -0,0 +1,128 @@
+using System.Text.RegularExpressions;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Time.Testing;
+using Moq;
+using SUI.StorageProcessFunction.Application;
+
+namespace Unit.Tests.StorageProcessFunction;
+
+public class StorageQueueMessageProcessorTests
```
jtaylor100, line 10: Nitpick (with my QA hat on): Ideas for some more test cases if you want the test to be more comprehensive:

- Queue message is valid, but file has been moved since a queue item was added. I.e. `blobFileReader.ReadAsync` fails to get the file.
- Cover the extra work you did in `BuildProcessedBlobName` to make it resilient to putting files under subdirectories, when generating the processed file name.
- `blobPayloadProcessor.ProcessAsync` errors
- `blobFileRedaer.ArchiveProcessedAsync` errors (Missing write permissions on an existing container or something)

The answer to some of these might just be 'let it fail' but worth considering. Feels like they could be written somewhere, maybe in a comment.

scripts/test-storage-process-function.ps1
```diff
@@ -0,0 +1,66 @@
+param(
```
jtaylor100, line 52: Ultra nitpick: Nice script! When I encountered it I was immediately interested in what it does or how to find out more info about it. I then found that when reading the `README.md` and the section on how to use it.

Powershell has a standard for documenting the info of scripts https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comment_based_help?view=powershell-7.5

Could it be useful to have some of the info, like prereqs for running, here instead of in the README? Or at the very least, a line saying 'See README.md for description and usage'.

jtaylor100 (changes requested):
As I mentioned, some could be bits to shelve for future PRs or work items. We can huddle tomorrow if you want to go through any of it together?

src/app-host/Program.cs
StuwiiDev: Really good shout, I was running it independently. However as it's always going to be needed with the new app with Matching running, I have added it into Aspire startup :) thanks

src/SUI.Client.StorageProcessFunction/Application/StorageQueueMessageProcessor.cs
StuwiiDev: good spot, I've updated it

src/SUI.Client.StorageProcessFunction/SUI.Client.StorageProcessFunction.csproj
StuwiiDev: though it does not effect build or anything beyond where the proj is sitting. I have moved it as it made sense :) thanks

src/SUI.Client/SUI.Client.StorageProcessFunction/Functions/ProcessStorageQueueMessageFunction.cs
StuwiiDev: It's called from the ProcessStorageQueueMessageFunction <- Where the queue trigger exists

src/SUI.Client/SUI.Client.StorageProcessFunction/Application/StorageQueueMessageProcessor.cs
StuwiiDev: done

src/SUI.Client/SUI.Client.StorageProcessFunction/README.md
StuwiiDev: This turned into a big update! Now importing official EventGrid schema package to ensure were using the correct one as per https://learn.microsoft.com/en-us/azure/event-grid/event-schema

scripts/test-storage-process-function.ps1
StuwiiDev: Great suggestion! had fun with this one

src/SUI.Client/SUI.Client.StorageProcessFunction/Application/BlobPayloadProcessor.cs
StuwiiDev: yea i broadly agree, my next PR does a lot of this sorting out 👍

src/SUI.Client/SUI.Client.StorageProcessFunction/Infrastructure/AzureStorage/AzureBlobFileReader.cs
StuwiiDev: by that Math i would expect files around 5MB which is well within a function apps capability of 1.5GB. However if we do start seeing large file sizes and seeing any degredation to the service then we can switch to streaming.

tests/Unit.Tests/StorageProcessFunction/StorageQueueMessageProcessorTests.cs
StuwiiDev: I'll add these test considerations to the next PR where I've done a fair bit of updating. This PR was just a quick skeleton

jtaylor100 (approved):

PR #302, opened 2026-04-01
Update CODEOWNERS

jtaylor100:
Think it makes sense, since @IgiCodes you're getting more involved.

The main practical thing is does is to auto-assign all of us for reviews. Though that doesn't mean we all need to review every PR, we can stick to the 'one review to merge' rule so we don't get bogged down.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @jtaylor100 @StuwiiDev
+* @jtaylor100 @StuwiiDev @IgiCodes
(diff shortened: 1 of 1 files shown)


IgiCodes:
Can I make a recommendation that you set up a Team in Github and reference that in the code owners file?
Saves you having to update the file all the time and also makes it easier to manage in general as you can add/remove people from the team a lot quicker in an emergency than needing to edit the file :slightly_smiling_face:

StuwiiDev:
> Can I make a recommendation that you set up a Team in Github and reference that in the code owners file? Saves you having to update the file all the time and also makes it easier to manage in general as you can add/remove people from the team a lot quicker in an emergency than needing to edit the file 🙂

I did start a team a while back sui-dev but never finished it or even used it but if you did want to go down this route, there is that

StuwiiDev (approved):

PR #303, opened 2026-04-02
docs: SUI-1644 - add Deployment IaC Stack Strategy design doc

IgiCodes:
## Summary
This PR adds a new design note capturing the proposed direction for restructuring deployment IaC so the repo can support multiple deployment architectures without breaking the current deployment path. The note formalises the stack-segregation approach for deployment infrastructure  and records the current assumptions, success criteria, and follow-on work for `SUI-1644`.

## Changes
- Adds a new design note at `documentation/docs/design/infra-stack-segregation.md`
- Documents the proposed direction for:
  - separating deployment IaC into architecture-based stacks
  - keeping generic stack/resource naming in a public repo
  - distinguishing shared platform infrastructure from architecture-specific stacks
  - treating application deployment as a separate layer consuming stack outputs
  - defining success criteria for preserving the current deployment path

## Validation
- Ran the MkDocs build successfully

### DIFF
--- a/documentation/docs/design/infra-stack-segregation.md
+++ b/documentation/docs/design/infra-stack-segregation.md
@@ -1,0 +1,255 @@
+# Deployment IaC Stack Strategy
+
+**Date:** `2026-04-02`
+
+**Scope:** Proposed direction for restructuring deployment IaC so different local-authority architectures can be deployed without breaking the current deployment path.
+
+This document captures the current proposed direction for deployment infrastructure in the SUI Matcher repo.
+
+It is a design note intended to:
+
+- formalise the current direction of travel
+- record the main structural changes needed in the existing IaC
+- make assumptions and follow-up work explicit
+
+---
+
+## 1. Overview
+
+The current deployment IaC is not structured to support multiple deployment architectures cleanly.
+
+Today the repo contains:
+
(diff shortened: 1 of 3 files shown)


documentation/docs/design/infra-stack-segregation.md
```diff
@@ -0,0 +1,255 @@
...
+
+This stack should expose stable outputs for downstream application deployment.
+
+### 4.2 Client agent stack
+
+Represents the current client-agent deployment shape:
+
+- shared platform resources
+- externally hosted client agent and related networking
+- firewall and routing components required for the current model
```
jtaylor100, line 102: Info (though you might already know): The 'as-is' firewall and routing is set up is differently deployed in the DfE infrastructure and the local authority's. 

In the DfE we spin up an Azure firewall whereas in the local authority we are peered to a hub network to reuse an existing firewall. The peering is not currently captured in the IaC -- don't think we have the permissions to manage it so had to request it, though was done before my time.

We're not planning right now to use hub networks in any other local authorities, so the environments should be consistent in those cases.

jtaylor100 (approved):
Looks good to me, just one point for info in case it throws a spanner in the works.

PR #305, opened 2026-04-08
feat: 1620 - processing pt2

StuwiiDev:
- Added storage-function blob orchestration for download, parse, process, and archive flow
- Replaced payload/queue processor abstractions with blob storage and orchestrator interfaces
- Added matching API client and wired HTTP matching into the storage function
- Added person-matching options, batch processing, and processing stats
- Added CSV parser factory, parser implementation, helpers, and record reader
- Updated matching/reconciliation file processors to use the new CSV reading flow
- Reorganized storage-function namespaces/files and renamed several core types
- Updated Aspire and local settings for matching API URL, parser selection, and processed container config
- Expanded the local test script with larger CSV test data
- Added and updated unit tests across storage processing, CSV parsing, matching, and orchestration

Out of scope of this PR:

- Dealing with results. This will be in another PR


src/SUI.Client/SUI.Client.StorageProcessFunction/Program.cs
```diff
@@ -29,10 +39,41 @@
...
+                    }
+
+                    client.BaseAddress = new Uri(options.MatchApiBaseAddress);
+                }
+            );
+
+            services.AddSingleton(
+                typeof(IMatchPersonRecordOrchestrator<>),
+                typeof(MatchPersonRecordOrchestrator<>)
+            );
```
jtaylor100, line 69: Trying to understand how this works. So a singleton is added which resolves IMatchPersonRecordOrchestrator to MatchPersonRecordOrchestrator. But it does this for all possible values of the generic type?

src/SUI.Client/SUI.Client.StorageProcessFunction/Program.cs
```diff
@@ -29,10 +39,41 @@
                 return new BlobServiceClient(connectionString);
             });
 
-            services.AddSingleton<IBlobFileReader, AzureBlobFileReader>();
-            services.AddSingleton<IBlobPayloadProcessor, BlobPayloadProcessor>();
-            services.AddSingleton<IStorageQueueMessageParser, EventGridStorageQueueMessageParser>();
-            services.AddSingleton<IStorageQueueMessageProcessor, StorageQueueMessageProcessor>();
+            services.AddSingleton<IBlobStorageClient, AzureBlobStorageClient>();
+            services.AddSingleton<IStorageQueueMessageParser, EventGridMessageParser>();
+            services.AddSingleton<IBlobFileOrchestrator, BlobFileOrchestrator>();
```
jtaylor100, line 44: Annoying question: Why put these classes behind interfaces? i.e. are any boundaries being crossed? Do we expect to swap them out? A design principle?

jtaylor100 (approved):
Looks good overall as an approach! Just two questions for my learning

src/SUI.Client/SUI.Client.StorageProcessFunction/Program.cs
StuwiiDev: yes so we don't have to specify each one individually. We could specify if we wanted to be strict

src/SUI.Client/SUI.Client.StorageProcessFunction/Program.cs
StuwiiDev: design principle for unit testability. Classes being unit tested with these classes can mock results

PR #308, opened 2026-04-15
refactor: 1629 console app job pt1

StuwiiDev:
- Removes old function app project
- Adds new console app project - skeleton at the moment. Next PR will add functionality for recieving jobs in queue
- Moved some tests around to make sense in the Unit.Tests project

Note: For Sonarcloud I had to remove old proj and add move tests around so it would pass


src/app-host/Program.cs
```diff
@@ -39,15 +39,10 @@
 // Feature flag setup
 var auditLoggingFlag = builder.Configuration.GetValue<string>("FeatureToggles:EnableAuditLogging");
 var auditLoggingEnabled = bool.Parse(auditLoggingFlag!);
-var storageProcessFunctionFlag = builder.Configuration.GetValue<bool>(
-    "FeatureToggles:EnableStorageProcessFunction"
-);
```
IgiCodes, line 44: **nitpick (non-blocking):** this removes the AppHost wiring for `EnableStorageProcessFunction`, but `src/app-host/appsettings.Development.json` still sets it to `true`. Could we clean that up here so local config matches current behavior?

IgiCodes (dismissed):
Looks good
Approving on the basis that this is an initial PR with a skeleton project and the knowledge that the project is currently not functional despite deleting the existing processor function 👍 

Left one minor non-blocking comment about some now dead config.

IgiCodes (approved):

PR #309, opened 2026-04-15
feat: 1620 - console app pt2 - processing storage queue message

StuwiiDev:
## Summary

This PR is to create the storage queue message reader. This involves reading messages, keeping them alive for the duration of the job and then handling success & failure of the job.

## Changes

- Azure Queue Integration - read and poison
- Queue Message Processing Implementation
- Configuration and Validation
- Application Startup and Lifecycle

## Validation

- Unit tests run
- Manual running of the application with the test script.

### DIFF
--- a/src/SUI.Client/SUI.Client.StorageProcessJob/Infrastructure/Azure/AzureStorageQueueClient.cs
+++ b/src/SUI.Client/SUI.Client.StorageProcessJob/Infrastructure/Azure/AzureStorageQueueClient.cs
@@ -24,4 +24,18 @@ public sealed class AzureStorageQueueClient(
             ? null
             : new StorageQueueMessage(message.MessageText, message.MessageId, message.PopReceipt);
     }
+
+    public async Task DeleteMessageAsync(
+        StorageQueueMessage message,
+        CancellationToken cancellationToken
+    )
+    {
+        var queueClient = queueServiceClient.GetQueueClient(options.Value.QueueName);
+
+        await queueClient.DeleteMessageAsync(
+            message.MessageId,
+            message.PopReceipt,
+            cancellationToken
+        );
+    }
 }
--- a/src/SUI.Client/SUI.Client.StorageProcessJob/Infrastructure/Azure/AzureStorageQueueClient.cs
+++ b/src/SUI.Client/SUI.Client.StorageProcessJob/Infrastructure/Azure/AzureStorageQueueClient.cs
@@ -1,5 +1,5 @@
-using System.Diagnostics.CodeAnalysis;
 using Azure.Storage.Queues;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using SUI.Client.StorageProcessJob.Application;
 using SUI.Client.StorageProcessJob.Application.Interfaces;
@@ -7,6 +7,7 @@ using SUI.Client.StorageProcessJob.Application.Interfaces;
 namespace SUI.Client.StorageProcessJob.Infrastructure.Azure;
 
 public sealed class AzureStorageQueueClient(
+    ILogger<AzureStorageQueueClient> logger,
     QueueServiceClient queueServiceClient,
     IOptions<StorageProcessJobOptions> options
 ) : IStorageQueueClient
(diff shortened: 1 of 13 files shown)


src/SUI.Client/SUI.Client.StorageProcessJob/Infrastructure/Azure/AzureStorageQueueClient.cs
```diff
@@ -0,0 +1,98 @@
...
+                cancellationToken: cancellationToken
+            );
+
+            await queueClient.DeleteMessageAsync(
+                message.MessageId,
+                message.PopReceipt,
+                cancellationToken
+            );
+
+            return null;
```
IgiCodes, line 53: **issue (blocking):** I think this path will fail the first time we try to poison a message. `GetQueueClient()` only gives us a client; it doesn’t create the queue. I couldn’t find any provisioning for `${QueueName}-poison` in the repo, so `SendMessageAsync(...)` will throw here and we’ll keep retrying the same bad message instead of sidelining it. Do we need to provision the poison queue explicitly or create it before sending?

src/SUI.Client/SUI.Client.StorageProcessJob/Infrastructure/QueueFileProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+        finally
+        {
+            await renewalCts.CancelAsync();
+            await IgnoreCancellationAsync(renewalTask); // Ensure task is finished before moving on
+        }
+
+        logger.LogInformation("Finished processing queue message {MessageId}.", message.MessageId);
+
+        // At this point we don't want to stop the message being deleted as it's been processed successfully.
+        await queueClient.DeleteMessageAsync(currentMessage, CancellationToken.None);
```
IgiCodes, line 60: **issue (blocking):** I think there’s a duplicate-processing risk here. If `blobFileOrchestrator.ProcessAsync(...)` succeeds but the background visibility-renewal task already thrown an exception, the `finally` block will surface that renewal failure and we never reach `DeleteMessageAsync(...)`. That means the message can remain in the queue even though the blob was already processed. Can we make sure a renewal failure doesn’t stop us deleting the message once the main processing path has completed successfully?

tests/Unit.Tests/Client/StorageProcessJob/QueueFileProcessorTests.cs
IgiCodes, line 1: **suggestion (if-minor):** Can we add a test for the renewal-failure path? The current tests cover successful renewal and successful deletion, but not the case where `RenewMessageVisibilityAsync(...)` fails while blob processing still succeeds. That looks like the highest-risk edge case in this flow.

src/SUI.Client/SUI.Client.StorageProcessJob/appsettings.Development.json
IgiCodes, line 1: **suggestion (non-blocking):** With the addition of azurite, do we not need to add `AzureWebJobsStorage="UseDevelopmentStorage=true"` to the dev appsettings?

IgiCodes (changes requested):

src/SUI.Client/SUI.Client.StorageProcessJob/Infrastructure/Azure/AzureStorageQueueClient.cs
StuwiiDev: yep good spot. Fixed

src/SUI.Client/SUI.Client.StorageProcessJob/Infrastructure/QueueFileProcessor.cs
StuwiiDev: quite edge case this one. queue sdk has retries so it would have to be quite a bad failure I reckon. In any case, I've fixed and added a test

tests/Unit.Tests/Client/StorageProcessJob/QueueFileProcessorTests.cs
StuwiiDev: test added

src/SUI.Client/SUI.Client.StorageProcessJob/appsettings.Development.json
StuwiiDev: it's at the top level of the file. I kept it inline with common structures

IgiCodes: Oh derp, my fault for not checking the full file and only relying on the PR diff :joy:

IgiCodes (approved):

PR #311, opened 2026-04-20
feat: 1620 - write successful results to success dir

StuwiiDev:
## Summary

Changes in this PR are to take the successfully matched results and write them to a CSV in a successful directory.

## Changes

- SuccessfulMatchRecord
- SuccessMatchFileWriter - main method that takes results, selects successful ones and sends to blob writer.
- UploadBlobAsync - Generic method to write to blob
- StorageProcessJobOptions and config - to add the success directory name

## Validation

- UInit tested.
- Manually run with real CSV file.

## Notes:

- No idempotency in file writing as we specifically do not allow retries on app level errors. We can re-visit this later if we decide to change
- **CSV headers are subject to change once we know what they will look like. Additionally there may be a PR soon to remove these hard coded headers and use configuration values instead.**
- any missing CSV headers are treated as a file level failure and nothing is processed.

### DIFF
--- a/src/SUI.Client/SUI.Client.StorageProcessJob/Application/BlobFileOrchestrator.cs
+++ b/src/SUI.Client/SUI.Client.StorageProcessJob/Application/BlobFileOrchestrator.cs
@@ -49,7 +49,6 @@ public sealed class BlobFileOrchestrator(
             cancellationToken
         );
 
-        // Just logging for now. Next: Process results
         logger.LogInformation(
             "Completed processing blob {BlobName}. Total records: {TotalRecords}, Successful matches: {SuccessfulMatches}.",
             queueMessage.BlobName,
--- a/src/SUI.Client/SUI.Client.StorageProcessJob/Application/BlobFileOrchestrator.cs
+++ b/src/SUI.Client/SUI.Client.StorageProcessJob/Application/BlobFileOrchestrator.cs
@@ -14,6 +14,7 @@ public sealed class BlobFileOrchestrator(
     TimeProvider timeProvider,
     IBlobStorageClient blobStorageClient,
     IMatchPersonRecordOrchestrator<CsvRecordDto> matchPersonRecordOrchestrator,
+    ISuccessMatchFileWriter successMatchFileWriter,
     IOptions<StorageProcessJobOptions> options
 ) : IBlobFileOrchestrator
 {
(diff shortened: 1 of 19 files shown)


src/SUI.Client/SUI.Client.StorageProcessJob/Application/SuccessMatchFileWriter.cs
```diff
@@ -0,0 +1,133 @@
...
+            if (successfulMatch is not null)
+            {
+                successfulMatches.Add(successfulMatch);
+            }
+        }
+
+        var csvContent = BuildCsv(successfulMatches);
+        var destinationBlobName = BuildSuccessBlobName(sourceBlobName);
+
+        await blobStorageClient.UploadBlobAsync(
```
IgiCodes: **question (if-minor):** Should we bail out here when `successfulMatches.Count == 0` instead of uploading a header-only file?

At the moment, I think this will still create a CSV in `success` when there are no high-confidence matches, or when every eligible row gets skipped for missing data. That makes blob existence in the `success` container look like “this file produced successful matches”, even though the output may contain zero data rows.

src/SUI.Client/SUI.Client.StorageProcessJob/Application/BlobFileOrchestrator.cs
```diff
@@ -48,14 +49,19 @@ CancellationToken cancellationToken
             cancellationToken
         );
 
-        // Just logging for now. Next: Process results
         logger.LogInformation(
             "Completed processing blob {BlobName}. Total records: {TotalRecords}, Successful matches: {SuccessfulMatches}.",
             queueMessage.BlobName,
             matchedResults.Count,
             matchedResults.Count(r => r.IsSuccess)
```
IgiCodes, line 56: **nitpick (if-minor):** Small mismatch here: this log reports `matchedResults.Count(r => r.IsSuccess)`, but the success file we write only includes records that are both `IsSuccess` and `IsHighConfidenceMatch`.

Do we want to change the number we log here or add an extra part of the log message to specify high confidence matches?

StuwiiDev: yes wording needs updating. It's more on how many did not Error as apposed to what is High Confidence. I'll change the log wording to 'Successful match calls' or something like that

src/SUI.Client/SUI.Client.StorageProcessJob/Application/SuccessMatchFileWriter.cs
StuwiiDev: I've updated to not write if there is no high confidence matches

src/SUI.Client/SUI.Client.StorageProcessJob/Application/BlobFileOrchestrator.cs
StuwiiDev: Updated to try and be more clear on log

IgiCodes (approved):

PR #313, opened 2026-04-22
feat: 1620 - move to config headers

StuwiiDev:
## Summary

Move from hard coded CSV header read and validation to configuration. The main advantage is not having multiple different parsers. Additional if deployed to another client then it's just a case of updating appsettings.json

## Changes
- Add column mappings to appsettings.json and Development
- Updated Csv validator to check CSV for configured headers
- Updated parser to use configured headers
- Added startup validation for a fail fast deployment
- Added a csv mapping provider interface so in the future when we do reconciliation we have a contract.

## Validation

- Unit tested
- Manually tested with test script on happy path and when a config is missing

### DIFF
--- a/src/SUI.Client/SUI.Client.StorageProcessJob/Application/BlobFileOrchestrator.cs
+++ b/src/SUI.Client/SUI.Client.StorageProcessJob/Application/BlobFileOrchestrator.cs
@@ -91,7 +91,7 @@ public sealed class BlobFileOrchestrator(
 
         if (missingOptionalHeaders.Length > 0)
         {
-            logger.LogInformation(
+            logger.LogWarning(
                 "Missing optional CSV headers: {MissingOptionalHeaders}. On {BlobName}",
                 string.Join(", ", missingOptionalHeaders),
                 queueMessage.BlobName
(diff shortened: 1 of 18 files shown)


src/SUI.Client/SUI.Client.StorageProcessJob/Application/BlobFileOrchestrator.cs
```diff
@@ -40,7 +41,10 @@ CancellationToken cancellationToken
 
         var blobStream = new StreamReader(blobContent.ToStream());
         var listOfRecords = await CsvRecordReader.ReadCsvTextAsync(blobStream, cancellationToken);
-        CsvHeaderValidator.Validate(listOfRecords.Headers, options.Value.CsvParserName);
+        CsvHeaderValidator.Validate(
+            listOfRecords.Headers,
+            csvRequiredHeadersProvider.GetRequiredHeaders()
+        );
```
IgiCodes: **question (if-minor):** Assuming `Email`, `Gender`, and `Phone` are intentionally not treated as required headers, should we still flag when one of those configured headers is missing from the file? At the moment it looks like we'd just quietly send less data to matching rather than fail or warn. If that is the intended behaviour, it might be worth adding a test for that path too so it's explicit.

StuwiiDev: I've added code to log missing optionals and added a test for it as well. It would be useful to be able to check logs for this kind of information

IgiCodes (approved):
LGTM ✅ 

Cheers for adding the additional validation/warning for the optional headers, solid addition. This is looking really robust now! 💪

PR #314, opened 2026-04-22
refactor: SUI-1679 - introduce IaC stack roots and shared modules

IgiCodes:
## Summary

Introduce deployable IaC stack roots under `infra/stacks` and extract the shared Bicep resources into reusable modules.

This adds a new full `client-agent` stack root and workflow for the DfE-hosted test topology, while preserving the existing `src/app-host/infra` deployment model. The legacy `src/SUI.Client/SUI.Client.Watcher/infra` path is left unchanged in this PR.

## Changes

- add shared Bicep modules under `infra/modules/shared` and a `client-agent` infrastructure module under `infra/modules/client-agent`
- add stack roots under `infra/stacks` for `client-agent` plus placeholder roots and README files for `blob-event-processor` and `api-batch-processor`
- refactor `src/app-host/infra` to compose the shared modules instead of carrying duplicate resource definitions
- add the full `infra/stacks/client-agent/main.bicep` root and `.github/workflows/gh-client-agent-infra-deploy.yml` for the DfE-hosted test stack
- keep `.github/workflows/gh-deploy-infra.yml` as the existing `src/app-host/infra` infrastructure workflow that still drives `azd provision`
- update the stack segregation design doc and infra deployment documentation to describe the new stack-root structure and how it relates to the existing app-host deployment flow

## Validation

- ran `az bicep build --file src/app-host/infra/main.bicep`
- ran `az bicep build --file infra/stacks/client-agent/main.bicep`
- ran `az bicep build --file src/SUI.Client/SUI.Client.Watcher/infra/client.bicep`
- compared the recursive deployed resource sets for `src/app-host/infra/main.bicep` on `main` vs this branch; the non-wrapper resource identities matched exactly (`15` vs `15`, with no resources only on one side)
- compared `what-if` output for `src/app-host/infra/main.bicep` on `main` vs this branch; the results matched exactly (`8 to modify, 2 no change, 49 to ignore`)
- validated the workflow YAML parses cleanly and `git diff --check` passes

## Follow-up

- `blob-event-processor` and `api-batch-processor` are placeholder stack roots only in this PR; runnable deployment workflows for those stacks should be added when their infrastructure becomes deployable


src/SUI.Client/SUI.Client.Watcher/infra/client.bicep
StuwiiDev, line 1: I am not entirely sure we need this file at all anymore, it came from a time before we used nuget packages for deployment and contains the Virtual Machine resource which is DfE specific

infra/modules/client-agent/infrastructure.bicep
```diff
@@ -0,0 +1,456 @@
...
+            id: vnet.properties.subnets[0].id
+          }
+          privateIPAllocationMethod: 'Dynamic'
+        }
+      }
+    ]
+  }
+}
+
+resource vm 'Microsoft.Compute/virtualMachines@2022-03-01' = {
```
StuwiiDev, line 110: VM was very specific to DfE deployment, this wouldn't be deployed to another unless specified. As it came from the Client.Watcher bicep I think it's quite out of date as well as we havnt used that since June 2025.

StuwiiDev (changes requested):
issue with VM in the client as discussed, comments revolve around that as well. I'll re-review after the vm update

src/SUI.Client/SUI.Client.Watcher/infra/client.bicep
StuwiiDev: flag has resolved this

StuwiiDev (approved):
I've checked manually on the 'legacy' depoy and it's working as expected as as before.

New stack has differences but we have already discussed that so I am happy with these changes 👍

PR #316, opened 2026-04-23
feat: 1620 - full file results output

StuwiiDev:
## Summary

This PR deals with outputting the full results to an output csv file. The purpose of this, is so we can analyse the files in a secure environment (Azure) without writing PII to logs.

## Changes

- Merged success file writer into a MatchResultsService to better represent result handling.
- Created ExportFullResultsAsync to handle writing all results to a file for later evaluation. This merges the matching API results into the original file data.
- Made a builder for file names to ensure the timestamp always matches for all file outputs - for convenience on searching files in explorer.
- Also removed the last remnance of the CsvParser type name 

## Validation

- Unit tested
- Manual test script run and checking outputs
-


src/SUI.Client/SUI.Client.StorageProcessJob/Application/MatchResultsService.cs
```diff
@@ -0,0 +1,227 @@
...
+    private const string CsvContentType = "text/csv";
+
+    public async Task ExportSuccessResultsAsync(
+        MatchResultsBlobNames blobNames,
+        string sourceBlobName,
+        IReadOnlyCollection<ProcessedMatchRecord<CsvRecordDto>> matchedResults,
+        CancellationToken cancellationToken
+    )
+    {
+        const string inputIdHeader = "Id";
```
IgiCodes: **issue (blocking):** This looks hardcoded back to `Id`, but the CSV schema is configurable now. If `ColumnMappings.Id` is set to something else, parsing/validation will pass, but the success export path below will skip every row as missing its ID. Should this read the configured ID mapping instead?

IgiCodes (changes requested):

src/SUI.Client/SUI.Client.StorageProcessJob/Application/MatchResultsService.cs
StuwiiDev: yes, and now updated

IgiCodes (approved):
LGTM ✅

PR #317, opened 2026-04-29
feat: 1621 - Add searchId to response

StuwiiDev:
## Summary

Adding the already existing SearchId to the response object. This gives a meaningful way to correlate a request from a client to the backing logging we produce.

## Changes

- Created a ActivityHashService and interface so we can inject the service. Interfaced for simpler unit testing and generally a better pattern.
- Removed HashUtil in favour of ActivityHashService
- Moved remaining Activity calls out of MatchingService and into the ActivityHashService.
- Updated MatchingService to use the ActivityHashService 
- Updated ReconcilationService to return with the matching services SearchId 
- Updated event procesor full output file to include searchId

## Validation

- All methods Unit tested
- All tests run
- Manually checked that the SearchId is returning when running a search


src/matching-api/Services/MatchingService.cs
```diff
@@ -111,13 +113,14 @@ PersonSpecificationForNoLogic personSpecification
                 Score = result.Result?.Score,
                 ProcessStage = result.ProcessStage,
             },
+            SearchId = activityHashService.GetUniqueSearchId(),
```
IgiCodes: **issue (blocking):** I think this endpoint will always return `searchId: null` at the moment, because `SearchNoLogicAsync` never stores a search id before reading it here.

Could we call `activityHashService.StoreUniqueSearchIdFor(personSpecification)` in this path too and return that value?

src/matching-api/Services/MatchingService.cs
```diff
@@ -95,6 +96,7 @@ await auditLogger.LogAsync(
                 ProcessStage = result.ProcessStage,
             },
             DataQuality = dataQualityResult,
+            SearchId = activityHashService.GetUniqueSearchId(),
```
IgiCodes: **suggestion (non-blocking):** As above, we could use the `searchId` local here.

src/matching-api/Services/MatchingService.cs
```diff
@@ -59,6 +59,7 @@ await auditLogger.LogAsync(
             {
                 Result = new MatchResult { MatchStatus = MatchStatus.Error },
                 DataQuality = dataQualityResult,
+                SearchId = activityHashService.GetUniqueSearchId(),
```
IgiCodes: **suggestion (non-blocking):** We already have the computed `searchId` local above, but this reads it back from `Activity.Current`. If there isn’t a current activity, the hash is still computed but this response value comes back as null. Could we return the local `searchId` instead?
I don't think this could realistically happen during a real request but making things more robust might be a good idea 🙂

IgiCodes (changes requested):
Couple of minor points

src/matching-api/Services/MatchingService.cs
StuwiiDev: this endpoint (or the one that calls this) isnt used to persist info but i can add it anyway. Seems like a value add

src/matching-api/Services/MatchingService.cs
StuwiiDev: Updated to use local searchId :)

IgiCodes (approved):

PR #319, opened 2026-04-29
feat: 1672 - remove unused audit feature

StuwiiDev:
## Summary

Removed the now redundent and unused audit feature. As the service is deployment per client and Machin2Machine

## Changes

- Removed from github CI
- Removed from Bicep
- Removed from code
- Removed the readme for audit

## Validation

- All tests passing. E2E, Int and unit
- Run a bicep what-if and no changes from current main branch


IgiCodes (changes requested):
**issue (blocking):** I think there’s one stale reference outside the files changed in this PR.

`.github/workflows/gh-deploy-infra.yml` still passes `enableAuditLogging` into `src/app-host/infra/main.bicep`, but this PR removes that bicep parameter. That will probably make the infra what-if/deploy workflow fail with an unknown parameter error.

There are also stale audit inputs/env vars in the same workflow, and `.github/workflows/gh-deploy-app.yml` still exposes `enable_audit_logging` / `AZURE_ENABLE_AUDIT_LOGGING`, which now looks unused. Worth tidying that up while we're here.

Might be worth just checking the whole repo for the removed parameters :)

IgiCodes (approved):

PR #321, opened 2026-04-30
feat: SUI-1698 - add blob-event-processor stack foundation

IgiCodes:
## Summary

Introduce the initial `blob-event-processor` stack foundation so the blob-event-processor architecture has a real deployable IaC root to build on.

This PR establishes the stack boundary, provisions the shared platform plus storage resources needed for the processing flow, and adds a manual workflow so the stack can be tested independently. It does not yet complete the Event Grid or ACA Job runtime wiring.

## Changes

- replace the `blob-event-processor` placeholder root with a real stack root that composes the shared identity, ACR, observability, ACA environment, secrets, and monitoring modules
- add a stack-specific storage module that provisions the storage account, `incoming` / `processed` / `success` blob containers, and the primary plus poison storage queues for the processing contract
- expose stack outputs for the shared platform and storage resources so follow-on deployment work can consume them
- add a manual GitHub Actions workflow to run what-if or deploy the `blob-event-processor` stack root directly
- update the infra/design documentation so it reflects the new stack boundary and the fact that this PR delivers the deployable foundation rather than the full runtime path

## Validation

- workflow YAML parsed successfully for `.github/workflows/gh-blob-event-processor-infra-deploy.yml`

## Follow-Up Context

This PR intentionally stops at the stack foundation. Event Grid wiring, ACA Job deployment of `SUI.Client.StorageProcessJob`, and any further networking/security hardening remain follow-up work.

### DIFF
--- a/.github/workflows/gh-blob-event-processor-infra-deploy.yml
+++ b/.github/workflows/gh-blob-event-processor-infra-deploy.yml
@@ -1,0 +1,91 @@
+name: Deploy Blob-Event-Processor Infrastructure
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: The environment target for deployment
+        required: true
+        default: 'Integration'
+      what_if:
+        description: Run bicep what-if to check for changes
+        default: true
+        type: boolean
+      version:
+        description: Deployment version, tag, branch, or SHA to deploy
+        required: true
+        type: string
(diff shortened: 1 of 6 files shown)


StuwiiDev (approved):
lgtm - can't spot anything out of place or any incorrect naming conventions. shall await to see result in next steps

PR #322, opened 2026-04-30
test: SUI-1703 - fix QueueFileProcessor test race

IgiCodes:
## Summary

This fixes the flaky `QueueFileProcessorTests` failure in `SUI.Client.StorageProcessJob`.

The failing test was relying on timing and was assuming the queue visibility renewal failure would always happen before processing completed. That wasn’t reliably true in CI, so this change makes the sequence explicit in the test without changing the production code.

## Changes

- update `Should_NotDeleteMessage_When_VisibilityRenewalFailsBeforeProcessingCompletes` so it no longer depends on timing
- replace the `Task.Delay`-based awaiting in the test harness with `TaskCompletionSource`
- use a mock `ILogger<QueueFileProcessor>` so the test can wait until the renewal failure has actually happened before letting processing complete
- review the other timing-sensitive tests in the same file and leave them as they are where the coordination already looks explicit

## Validation

- `dotnet test tests/Unit.Tests/Unit.Tests.csproj --filter FullyQualifiedName~Unit.Tests.Client.StorageProcessJob.QueueFileProcessorTests`
- ran the targeted `QueueFileProcessorTests` 5 times in a row successfully

## Context

CI was failing with:
- `Unit.Tests.Client.StorageProcessJob.QueueFileProcessorTests.Should_NotDeleteMessage_When_VisibilityRenewalFailsBeforeProcessingCompletes [FAIL]`
- `Assert.Throws() Failure: No exception was thrown`

### DIFF
--- a/tests/Unit.Tests/Client/StorageProcessJob/QueueFileProcessorTests.cs
+++ b/tests/Unit.Tests/Client/StorageProcessJob/QueueFileProcessorTests.cs
@@ -1,4 +1,4 @@
-using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Time.Testing;
 using Moq;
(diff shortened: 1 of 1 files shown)


robshakespeare (approved):
Smart :+1:  Thanks for fixing this, always good to fix flakey tests 🤩

PR #323, opened 2026-05-01
docs: stategy 5 update

StuwiiDev:
## Summary

Added 3 new sections to the strategies.md on stategy 5. These are to attempt to explain why we started it, how we went about it and then what we got as a result.

## Changes

- Updated strategies.md

## Validation

- A peer review to ensure it reads well and makes sense.


jtaylor100 (approved):
Aced it 

Image

PR #326, opened 2026-05-07
feat: SUI-1712 - add stack-owned resource group wrappers

IgiCodes:
## Summary
This moves the stack deployment model away from shared, pre-created resource groups and toward stack-owned ones.

The aim here is to make each stack responsible for the resource group it lives in, so deployments are cleaner and future teardown/decommission work can be handled at the stack level rather than by manually untangling shared environments.

## Changes

- added subscription scope wrappers for `client-agent`, `blob-event-processor`, and `api-batch-processor` so each stack can create or update its own resource group before deploying the existing stack root
- updated the `client-agent` and `blob-event-processor` workflows to run subscription scoped what-if/deploy against those wrappers instead of targeting a pre-existing shared resource group
- updated the stack and design docs so the expected contract is now that each stack owns its own resource group

## Validation

- ran `az bicep build` for the new subscription wrappers
- checked that the updated workflow YAML parses

### DIFF
--- a/infra/stacks/blob-event-processor/subscription.bicep
+++ b/infra/stacks/blob-event-processor/subscription.bicep
@@ -1,0 +1,93 @@
+targetScope = 'subscription'
+
+@minLength(1)
+@description('Name of the deployment environment used for stack resource naming')
+param environmentName string
+
+@minLength(1)
+@description('The prefix used for all deployed resources and the resource-group naming convention')
+param environmentPrefix string
+
+@minLength(1)
+@description('The version number of the container app managed environment, used for naming convention')
+param containerAppManagedEnvironmentNumber string
+
+@minLength(1)
+@description('The address prefix for the virtual network')
+param containerAppVnet string
+
+@minLength(1)
+@description('Container App environment subnet')
+param containerAppEnvSubnet string
+
+@minLength(1)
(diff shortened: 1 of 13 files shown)


infra/stacks/blob-event-processor/subscription.bicep
```diff
@@ -0,0 +1,93 @@
...
+    environmentPrefix: environmentPrefix
+    containerAppManagedEnvironmentNumber: containerAppManagedEnvironmentNumber
+    containerAppVnet: containerAppVnet
+    containerAppEnvSubnet: containerAppEnvSubnet
+    location: location
+    monitoringActionGroupEmail: monitoringActionGroupEmail
+    turnOnAlerts: turnOnAlerts
+  }
+}
+
```
StuwiiDev, line 65: from the stackDeployment, do we need to emit the EventGrid subscription and topic?
if not feel free to close this comment

IgiCodes: They aren't used anywhere yet as far as I can see... But I’ve added them just so the subscription level wrapper stays in sync with the underlying stack outputs rather than hiding the Event Grid bits if/when we want to consume them later 👌

StuwiiDev (approved):

PR #331, opened 2026-05-19
fix: SUI-1761 - harden stack-owned IaC deployment flow

IgiCodes:
## Summary

This PR hardens the stack-owned IaC flow for the blob-event-processor and client-agent stacks so they can be deployed, redeployed, and moved between allowed regions without colliding with existing shared resource names or Azure deployment metadata.

## Changes

- add stack-specific suffixes to the shared infrastructure modules and wire them through the blob-event-processor and client-agent stack entrypoints
- make workflow region selection come from `vars.AZURE_LOCATION`, including the smoke test workflow callers
- include the Azure region in subscription-scope `what-if` and `deploy` names to avoid `InvalidDeploymentLocation` clashes when switching regions
- shorten constrained resource names and add deterministic uniqueness where needed

## Validation

- `az bicep build --file infra/stacks/blob-event-processor/subscription.bicep`
- `az bicep build --file infra/stacks/client-agent/main.bicep`
- `az bicep build --file infra/stacks/client-agent/subscription.bicep`
- successful `gh-blob-event-processor-infra-deploy.yml` deployment after applying the naming fixes
- full resource-group delete and redeploy


infra/modules/shared/container-registry.bicep
```diff
@@ -7,12 +7,17 @@ param environmentPrefix string
 @description('The lowercase environment name used for resource naming')
 param lowercaseEnvironmentName string
 
+@description('Short stack-specific suffix used to avoid cross-stack name collisions.')
+param stackNameSuffix string = ''
+
 @description('Tags that will be applied to all resources')
 param tags object = {}
 
+var stackNameToken = empty(stackNameSuffix) ? '' : toLower(stackNameSuffix)
```
StuwiiDev, line 16: question: will this cause any naming conflicts if 2 stacks deployed and no suffix was entered?

IgiCodes: Yeah, if someone deliberately doesn’t pass a suffix then they'd be opting back into the old shared naming pattern, so a collision would be expected. For the stack-owned flows in this PR we're explicitly passing suffixes (`bep` / `ca`), which is what avoids the clashes here :slightly_smiling_face: 
We could change it to *require* a suffix I guess but I figured having the flexibility to not provide one would be better.

StuwiiDev (approved):

PR #337, opened 2026-05-26
fix: Add 'include role assignments' parameter

jtaylor100:
Required to account for cloud environments where access to assign roles is restricted. Roles can be added manually or through due process.

Tried to find something that could 'import' the assignments back into the bicep deployments, but no such thing exists.


robshakespeare (approved):
Looks good.  A sensible option provided over the top given the constraints.

PR #341, opened 2026-05-28
feat: SUI-1973 - update blob stack to support existing resources

IgiCodes:
## Summary

Updates the blob-event-processor stack so it can deploy either fully stack-owned resources or integrate with client-provided resources. This keeps the current default create path, while adding an existing resource group / existing storage account mode for client environments.

## Changes

- Adds resource group and storage account modes using `create` / `existing` values.
- Updates the storage module to create our standard containers, queues, blob/queue private endpoints, and private DNS wiring against either a created or existing storage account.
- Wires the new parameters through the blob stack subscription wrapper, stack root, GitHub workflow, and shared stack deploy action.
- Documents direct `az deployment sub` usage for environments deployed from an approved device rather than GitHub Actions.

## Validation

- `az bicep build` of each file
- yaml parser checks

## Notes

- Existing storage mode still uses our standard names: `incoming`, `processed`, `success`, `storage-process-job`, and `storage-process-job-poison`.
- Existing resource group deployments are outside the stack-owned resource group teardown contract and need selective cleanup.
- Need to validate these changes work by spinning up "existing resources" in an Azure RG to simulate the client infra

### DIFF
--- a/.github/actions/stack-infra-run/action.yml
+++ b/.github/actions/stack-infra-run/action.yml
@@ -72,6 +72,22 @@ inputs:
     description: Whether or not to include role assignments, since some environments may restrict these. 
     default: true
     type: boolean
+  resource_group_mode:
+    description: "Optional resource group mode: create or existing"
+    required: false
+    default: ""
+  target_resource_group_name:
+    description: "Optional existing resource group name when resource_group_mode is existing"
+    required: false
+    default: ""
+  storage_account_mode:
+    description: "Optional storage account mode: create or existing"
+    required: false
+    default: ""
+  existing_storage_account_name:
+    description: "Optional existing storage account name when storage_account_mode is existing"
+    required: false
+    default: ""
 
 
 runs:
@@ -137,6 +153,22 @@ runs:
 
         if [ -n "${{ inputs.storage_process_job_image_tag }}" ]; then
           PARAMETERS+=("storageProcessJobImageTag=${{ inputs.storage_process_job_image_tag }}")
+        fi
+
+        if [ -n "${{ inputs.resource_group_mode }}" ]; then
+          PARAMETERS+=("resourceGroupMode=${{ inputs.resource_group_mode }}")
+        fi
+
+        if [ -n "${{ inputs.target_resource_group_name }}" ]; then
+          PARAMETERS+=("targetResourceGroupName=${{ inputs.target_resource_group_name }}")
+        fi
+
+        if [ -n "${{ inputs.storage_account_mode }}" ]; then
+          PARAMETERS+=("storageAccountMode=${{ inputs.storage_account_mode }}")
+        fi
+
+        if [ -n "${{ inputs.existing_storage_account_name }}" ]; then
+          PARAMETERS+=("existingStorageAccountName=${{ inputs.existing_storage_account_name }}")
         fi
 
         cleanup() {
--- a/.github/workflows/gh-blob-event-processor-infra-deploy.yml
+++ b/.github/workflows/gh-blob-event-processor-infra-deploy.yml
@@ -35,6 +35,28 @@ on:
         description: Include role assignments? # since some environments may restrict these.
         default: true
         type: boolean
+      resource_group_mode:
+        description: Resource group mode
+        default: create
+        type: choice
+        options:
+          - create
+          - existing
+      target_resource_group_name:
+        description: Existing resource group name when resource_group_mode is existing
+        required: false
+        type: string
+      storage_account_mode:
+        description: Storage account mode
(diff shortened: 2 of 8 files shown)


infra/modules/blob-event-processor/storage.bicep
```diff
@@ -45,9 +57,15 @@ resource storageAccount 'Microsoft.Storage/storageAccounts@2023-05-01' = {
   }
 }
 
+resource existingStorageAccount 'Microsoft.Storage/storageAccounts@2023-05-01' existing = if (storageAccountMode == 'existing') {
+  name: existingStorageAccountName
+}
+
 resource blobService 'Microsoft.Storage/storageAccounts/blobServices@2023-05-01' = {
```
robshakespeare: **question:** Good set of changes in this PR, looks like a tricky problem nicely solved.

I was just wondering, should things like the `blobService` and `queueService` have the `if (storageAccountMode == 'create')` too?

infra/modules/blob-event-processor/storage.bicep
```diff
@@ -45,9 +57,15 @@ resource storageAccount 'Microsoft.Storage/storageAccounts@2023-05-01' = {
...
-  parent: storageAccount
-  name: 'default'
+  name: '${storageAccountName}/default'
+  dependsOn: [
```
robshakespeare: **suggestion:** I think this should strictly stay with using 'parent' instead of 'dependsOn' because its a child resource: https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter-rule-use-parent-property

Maybe with the inclusion of the `if` conditional which I think is also needed, this will then work.

Something like:

```
resource blobService 'Microsoft.Storage/storageAccounts/blobServices@2023-05-01' = if (storageAccountMode == 'create') {
  parent: createdStorageAccount
  name: '${storageAccountName}/default'
}
```

infra/modules/blob-event-processor/storage.bicep
IgiCodes: We're intentionally still creating the blob/queue when the storage account already exists :slightly_smiling_face: 

Appreciate the callout to check though!

infra/modules/blob-event-processor/storage.bicep
IgiCodes: Agree on preferring `parent` where we can. The awkward bit here is that the storage account parent is conditional... in create mode it is `createdStorageAccount`, and in existing mode it is `existingStorageAccount`.

I do not think the `blobService` / `queueService` resources should be create-only, because existing storage mode still needs them so we can create our containers and queues in the supplied account.

I think the options are either:

1. keep this as-is with the explicit name and dependency, or
2. split the blob/queue service and child resources into separate create/existing branches so each branch can use `parent`.

I leaned towards the current version to avoid duplicating the containers and queues, but happy to split it if you think keeping `parent` throughout is worth the extra verbosity.

IgiCodes: I would suggest we do something like:
```
parent: storageAccountMode == 'create' ? createdStorageAccount : existingStorageAccount
```

But, `parent` needs to be a compile time value from what I've read. So, sadly that wouldn't work.

IgiCodes: Actually, you have made me realise that the `dependsOn` value is only `createdStorageAccount`, even when we're in `existing` mode. Let me have another look at this and refactor things a bit.

infra/modules/blob-event-processor/storage.bicep
robshakespeare: All good, thank you 🙂  I hadn't realise that was the intention.

infra/modules/blob-event-processor/storage.bicep
robshakespeare: Ok thanks.  Thanks for the additional context and information.  Now I understand more about this, I was going to say I agree let's keep this as-is.  This is a good solution.

But if there's changes that need to be done, that's cool :+1:

IgiCodes: I split the storage Bicep out so the create/existing storage account paths are clearer, with the shared bits pulled into a common module. Should be a bit easier to mentally parse now :ok_hand:

robshakespeare (commented):
Nice, thanks.  That flows well now, and is easier to understand how it works :+1:

infra/modules/blob-event-processor/storage.bicep
robshakespeare: Thanks, yes much easier to follow now :+1:

jtaylor100 (approved):

PR #344, opened 2026-06-01
feat: 1762 - firewall and peering

StuwiiDev:
## Summary

Add a firewall, route table and peering so traffic in the CAE VNET can only pass out through it.

## Changes

- CAE now has a routeTableId paramter that is required
- Created the egress Firewall and policies as shared module.
- Created Peering as shared module
- Created a Public IP for the firewall (Unsure if this is possible just yet)
- 

## Validation

- run local bicep file checks
- run a what-if https://github.com/DFE-Digital/SUI_Matcher/actions/runs/26813119847

### DIFF
--- a/infra/modules/shared/egress-firewall.bicep
+++ b/infra/modules/shared/egress-firewall.bicep
@@ -1,0 +1,243 @@
+@description('The location used for all deployed resources')
+param location string
+
+@description('Name of the deployment environment')
+param environmentName string
+
+@description('The prefix used for all deployed resources')
+param environmentPrefix string
+
+@description('Short stack-specific suffix used to avoid cross-stack name collisions.')
+param stackNameSuffix string = ''
+
+@description('The login server for the container registry to allow through the firewall')
+param containerRegistryEndpoint string
+
+@description('The name of the shared Key Vault')
+param keyVaultName string
+
+@description('The Key Vault FQDN allowed through the firewall')
+#disable-next-line no-hardcoded-env-urls
+param keyVaultEndpoint string = '${keyVaultName}.vault.azure.net'
+
+@description('The address prefix for the firewall virtual network')
+param firewallVnetAddressPrefix string = '192.168.4.0/23'
+
+@description('The address prefix for the Azure Firewall subnet')
+param firewallSubnetAddressPrefix string = '192.168.4.0/25'
+
+@description('The address spaces of the CAE virtual network (used as source address ranges in the firewall policy)')
+param caeVnetAddressPrefixes array
+
+@description('Tags that will be applied to all resources')
+param tags object = {}
+
+var lowercaseEnvironmentName = toLower(environmentName)
+var stackNameToken = empty(stackNameSuffix) ? '' : '-${toLower(stackNameSuffix)}'
+
+var firewallVnetName = '${environmentPrefix}-${lowercaseEnvironmentName}${stackNameToken}-vnetfw-01'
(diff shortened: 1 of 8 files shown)


infra/modules/shared/egress-firewall.bicep
```diff
@@ -0,0 +1,366 @@
...
+    // Basic-tier ACR serves image layers from shared Azure Storage; narrowing
+    // this requires upgrading ACR to Premium with dedicated data endpoints.
+    name: 'acr-blob-allow'
+    #disable-next-line no-hardcoded-env-urls
+    fqdn: '*.blob.core.windows.net'
+  }
+  {
+    name: 'kv-allow'
+    fqdn: keyVaultEndpoint
+  }
```
jtaylor100: Won't these be covered by the private endpoints?

infra/modules/shared/egress-firewall.bicep
```diff
@@ -0,0 +1,366 @@
...
+    fqdn: '*.identity.azure.net'
+  }
+  {
+    name: 'aspire-dashboard-allow'
+    #disable-next-line no-hardcoded-env-urls
+    fqdn: '${containerAppRegion}.ext.azurecontainerapps.dev'
+  }
+]
+
+resource firewallVirtualNetwork 'Microsoft.Network/virtualNetworks@2024-05-01' = {
```
jtaylor100, line 127: Does it need to be in a separate virtual network to be peered, or just a different subnet in the same network? I assumed the latter, but not sure if you've hit any technical bump

StuwiiDev: 🤔 it's the way we did it in the current live deployment, Firewall sits in a seperate VNET as it's own thing and then we peer them, otherwise it would need to go into the CAE VNET coupled together - I'll need to check pros/cons of both ways

jtaylor100: I think that was just done because the firewall was managed by a different team. Whereas in the new deployment we will manage the firewall ourselves.

It could be a better design to keep it in a different network, allow the deployment to more easily be done for these two cases. Although I'm not sure if extra costs are incurred by the peering/network.

infra/modules/shared/egress-firewall.bicep
StuwiiDev: keyvault will and can be removed from here.
ACR is not behind a private link so it does need to be here. + basic tier acr doesnt support private links

infra/modules/shared/egress-firewall.bicep
StuwiiDev: https://azure.microsoft.com/en-us/pricing/details/virtual-network/ - free to use VNETS and peering.
cost is in the data transfer which is 0.008 per GB

infra/modules/shared/egress-firewall.bicep
jtaylor100: Ah, but don't we have a private link to blob storage? Which is what this domain is pointed to

infra/modules/shared/egress-firewall.bicep
jtaylor100: I think proceed with separate VNETs and peering in that case, costs will be low 👍🏻

infra/modules/shared/egress-firewall.bicep
StuwiiDev: This has now been updated to use data endpoints host names from the ACR premium output

jtaylor100 (approved):

PR #351, opened 2026-06-04
feat: Upgrade ACR to premium

StuwiiDev:
## Summary

Moving ACR to premium tier so that we use data endpoints. This is a security measure to avoid exfiltration attack vectors to other blob storage accounts

## Changes

- Updated blob stack
- Updated client stack
- Upgraded ACR to premium

## Validation

- Ran az build on blob, client and container registry
- https://github.com/DFE-Digital/SUI_Matcher/actions/runs/26943391758 - Shows SKU going from Basic to Premium and shows dataEndpointEnabled = true


jtaylor100 (commented):
Won't you need to whitelist the 'data endpoint host names' value in the firewall? So that only that endpoint can be used, instead of *.blob

jtaylor100 (approved):

PR #357, opened 2026-06-10
fix: SUI-1973 - support existing storage resources

IgiCodes:
## Summary

Adds safer support for the blob-event-processor stack when using a pre-existing storage account, while reducing expected what-if noise from Azure-managed defaults and DfE policy tags.

## Changes

- Treat storage account `blobServices/default` and `queueServices/default` as existing service handles, so the stack only manages the SUI-owned containers, queues, private endpoints, DNS, and Event Grid wiring.
- Preserve default blob container properties explicitly to avoid repeated container drift in what-if output.
- Add optional tag controls for policy-shaped environments:
  - `tagEnvironmentName` lets deployments align the `Environment` tag with policy expectations.
  - `additionalTags` / `AZURE_ADDITIONAL_TAGS` lets DfE-specific tags such as `Service Offering` be supplied from repo variables instead of being hard-coded into client infrastructure.
- Pass the shared stack tags into Key Vault without replacing its required Aspire tag.
- Document existing-storage behaviour and the optional tag configuration for GitHub Actions and direct deployments.

## Validation

- Built `infra/stacks/blob-event-processor/main.bicep` with `az bicep build`.
- Built `infra/stacks/blob-event-processor/subscription.bicep` with `az bicep build`.
- Ran create-mode what-if from the branch and confirmed the storage account, containers, queues, private endpoints, and Event Grid resources are no-change.

### DIFF
--- a/infra/modules/blob-event-processor/storage-resources.bicep
+++ b/infra/modules/blob-event-processor/storage-resources.bicep
@@ -40,6 +40,8 @@ resource containers 'Microsoft.Storage/storageAccounts/blobServices/containers@2
     parent: blobService
     name: containerName
     properties: {
+      defaultEncryptionScope: '$account-encryption-key'
+      denyEncryptionScopeOverride: false
       publicAccess: 'None'
     }
   }
--- a/infra/modules/blob-event-processor/storage-resources.bicep
+++ b/infra/modules/blob-event-processor/storage-resources.bicep
@@ -31,7 +31,7 @@ var queueNames = [
   poisonQueueName
 ]
 
-resource blobService 'Microsoft.Storage/storageAccounts/blobServices@2023-05-01' = {
+resource blobService 'Microsoft.Storage/storageAccounts/blobServices@2023-05-01' existing = {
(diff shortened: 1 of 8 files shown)


infra/modules/blob-event-processor/storage-resources.bicep
```diff
@@ -40,12 +40,14 @@ resource containers 'Microsoft.Storage/storageAccounts/blobServices/containers@2
     parent: blobService
     name: containerName
     properties: {
+      defaultEncryptionScope: '$account-encryption-key'
```
StuwiiDev, line 43: does $account-encryption-key need to be wrapped in {}? if not close this comment

IgiCodes: This should stay as the literal `$account-encryption-key`. `${...}` would make it a Bicep interpolation, but here we want Azure Storage’s built-in default encryption scope name :slightly_smiling_face:

StuwiiDev (approved):

PR #359, opened 2026-06-11
docs: blob process local runbook

StuwiiDev:
## Summary

Adds a concise local runbook for the blob process so contributors can run and verify it consistently outside deployed environments. I've attempted to make it usable for both a new deploy and repeat deployments.
It does require being updated along with any workflow changes but the workflows have been mentioned in the doc as well

Fixed some issues found when going through the runbook.

## Changes

- Added documentation for prerequisites and local setup for the blob process.
- Documented step-by-step local execution workflow.
- Added guidance on expected results and basic troubleshooting checks.
- Updated bicep code to handle hyphen removal where there should be none.

## Validation

- Reviewed runbook steps for accuracy and readability.
- Confirmed instructions are complete enough for local execution without extra context.
- The full flow has been followed and worked successfully

### DIFF
--- a/documentation/docs/runbooks/blob-processor-local-runbook.md
+++ b/documentation/docs/runbooks/blob-processor-local-runbook.md
@@ -16,23 +16,31 @@ TODO:
 
 ## Deployment steps overview
 
-1. Pick a branch, tag or commit to deploy.
-2. Run dotnet restore, build and run all tests to ensure the code is in a good state.
-3. Run the infrastructure what-if to validate the output.
-4. Run the infrastructure deployment which will first deploy the infrastructure and then deploy the application to the environment.
+1. [Pick a branch, tag or commit to deploy](#pick-a-branch-tag-or-commit-to-deploy).
+2. [Run dotnet restore, build and run all tests to ensure the code is in a good state](#run-dotnet-restore-build-and-tests).
+3. [Set the deployment environment variables](#set-the-deployment-environment-variables).
+4. [Run the infrastructure what-if to validate the output](#run-the-infrastructure-what-if-to-validate-the-output).
+5. [Run the infrastructure deployment](#run-the-infrastructure-deploy).
 
     **Note** this will fail on first run for the applications as they will not yet exist in the azure container registry,
 but this is expected. The second run will succeed as the applications will have been built and pushed to the registry by the first run.
 
-5. Run the API' and Storage processor deployment which will deploy the API and Storage processor applications to the Azure container registry.
-6. Run the infrastructure deployment again to deploy the applications to the environment.
-7. Run the smoke tests to validate the deployment. You can do this by placing a csv file in the blob storage and checking the logs of the storage processor to see if it has processed the file.
+6. Run the API' and Storage processor deployment which will deploy the API and Storage processor applications to the Azure container registry.
+7. Run the infrastructure deployment again to deploy the applications to the environment.
+
+    **Note** this will succeed as the applications will now exist in the azure container registry.
+8. Run the smoke tests to validate the deployment. You can do this by placing a csv file in the blob storage and checking the logs of the storage processor to see if it has processed the file.
 
 ## Detailed deployment steps - Follow in order
 
-### Pick a branch, tag or commit to deploy.
+### Pick a branch, tag or commit to deploy. Ensure the branch is clean of any uncommited changes. Example for main:
 
-### Run dotnet restore, build and run all tests to ensure the code is in a good state.
+```bash
+git checkout main
+git pull
+```  
+
+### Run dotnet restore, build and tests
 
 ```bash
 dotnet restore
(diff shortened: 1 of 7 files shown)


infra/modules/shared/container-registry.bicep
```diff
@@ -14,10 +14,11 @@ param stackNameSuffix string = ''
 param tags object = {}
 
 var stackNameToken = empty(stackNameSuffix) ? '' : toLower(stackNameSuffix)
+var noHypenEnvironment = replace(environmentPrefix, '-', '')
 
 // The resource name can only contain alphanumeric characters.
 resource containerRegistry 'Microsoft.ContainerRegistry/registries@2023-07-01' = {
-  name: '${environmentPrefix}${lowercaseEnvironmentName}${stackNameToken}acr01'
+  name: '${noHypenEnvironment}${lowercaseEnvironmentName}${stackNameToken}acr01'
```
IgiCodes, line 21: **issue (blocking):** Do the image workflows need the same ACR name sanitising?
This PR updates the Bicep to strip hyphens from `environmentPrefix` when creating the registry, and the runbook does the same before running `az acr ...`.

But the existing image workflows (`gh-blob-event-processor-api-images.yml` and `gh-blob-event-processor-storage-job-image.yml`) still build `ACR_NAME` from the raw prefix: `ACR_NAME="${AZURE_ENV_PREFIX}${LOWERCASE_ENVIRONMENT_NAME}bepacr01"`

So if the prefix has a hyphen, infra creates one ACR name and the publish workflows look for a different one?

documentation/docs/runbooks/blob-processor-local-runbook.md
```diff
@@ -0,0 +1,516 @@
...
+
+# Workflow defaults. Change these only when the target deployment requires it.
+AZURE_INCLUDE_ROLE_ASSIGNMENTS="true"
+AZURE_TURN_ON_ALERTS="false"
+STORAGE_PROCESS_JOB_IMAGE_TAG="latest"
+MATCHING_API_IMAGE_TAG="latest"
+EXTERNAL_API_IMAGE_TAG="latest"
+STORAGE_ACCOUNT_MODE="create" # "create" or "existing"
+EXISTING_STORAGE_ACCOUNT_NAME="" # Leave blank if it does not exist.
+```
```
IgiCodes, line 108: **suggestion (non-blocking):** The environment we're deploying to probably doesn’t need extra Azure tags or an `Environment` tag override, but it might still be worth mentioning the optional env vars for `AZURE_TAG_ENVIRONMENT_NAME` / `AZURE_ADDITIONAL_TAGS` here.

Not a blocker, just keeps the local runbook aligned with the supported `tagEnvironmentName` / `additionalTags` parameters 👌

IgiCodes (changes requested):

infra/modules/shared/container-registry.bicep
StuwiiDev: valid and updated :)

documentation/docs/runbooks/blob-processor-local-runbook.md
StuwiiDev: I've added those in as optional

documentation/docs/runbooks/blob-processor-local-runbook.md
```diff
@@ -226,7 +232,9 @@ az deployment group what-if \
     matchingApiImageTag="${MATCHING_API_IMAGE_TAG}" \
     externalApiImageTag="${EXTERNAL_API_IMAGE_TAG}" \
     storageAccountMode="${STORAGE_ACCOUNT_MODE}" \
-    existingStorageAccountName="${EXISTING_STORAGE_ACCOUNT_NAME}"
+    existingStorageAccountName="${EXISTING_STORAGE_ACCOUNT_NAME}" \
+    tagEnvironmentName="${AZURE_TAG_ENVIRONMENT_NAME:-}" \
+    additionalTags="${AZURE_ADDITIONAL_TAGS:-{}}"
```
IgiCodes: **issue (blocking):** I think this won't work the way you want it to. Bash will interpret the first `}` as the end of the expression. You'll need to quote the `{}` for this to work. (i.e. `additionalTags="${AZURE_ADDITIONAL_TAGS:-"{}"}"`

However, the `.env` template further up already sets a default so if someone is following the runbook you shouldn't need to reset a default here and you can just do `additionalTags="${AZURE_ADDITIONAL_TAGS}"` 🙂

IgiCodes (changes requested):

documentation/docs/runbooks/blob-processor-local-runbook.md
StuwiiDev: updated to just remove the fallback 👍

IgiCodes (approved):
LGTM ✅

PR #360, opened 2026-06-16
use dotnets ImportFromPEM to import RSA

StuwiiDev:
## Summary

There is a simpler way to use RSA keys with PEM, and that is a ready built method in the RSA lib `ImportFromPem` this automatically handles PEM keys. This is a marked improvement from what we had where we were manually stripping out the begin/end which is error prone if it isnt exact

## Changes

- Updated PEM method in JwtHandler to use `ImportFromPem`

## Validation

- All tests pass
- Locally ran with my own key
- Tested in a deployed environment as well

### DIFF
--- a/src/external-api/Util/JwtHandler.cs
+++ b/src/external-api/Util/JwtHandler.cs
@@ -53,12 +53,8 @@ public class JwtHandler : IJwtHandler
 
     private static SigningCredentials FromPrivateKey(string privateKey, string kid)
     {
-        privateKey = privateKey.Replace("-----BEGIN RSA PRIVATE KEY-----", "");
-        privateKey = privateKey.Replace("-----END RSA PRIVATE KEY-----", "");
-        var keyBytes = Convert.FromBase64String(privateKey);
-
         var rsa = RSA.Create();
-        rsa.ImportRSAPrivateKey(keyBytes, out _);
+        rsa.ImportFromPem(privateKey.ToCharArray());
 
         var rsaSecurityKey = new RsaSecurityKey(rsa) { KeyId = kid };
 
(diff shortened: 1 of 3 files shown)


robshakespeare (dismissed):

src/external-api/Util/JwtHandler.cs
```diff
@@ -53,12 +53,8 @@ public string GenerateJwt(
 
     private static SigningCredentials FromPrivateKey(string privateKey, string kid)
     {
-        privateKey = privateKey.Replace("-----BEGIN RSA PRIVATE KEY-----", "");
-        privateKey = privateKey.Replace("-----END RSA PRIVATE KEY-----", "");
-        var keyBytes = Convert.FromBase64String(privateKey);
-
         var rsa = RSA.Create();
-        rsa.ImportRSAPrivateKey(keyBytes, out _);
+        rsa.ImportFromPem(privateKey.ToCharArray());
```
IgiCodes, line 57: **issue (if-minor):** I think this breaks the dev `StubTokenService` fallback in `src/external-api/Services/StubTokenService.cs`. When `NHS_DIGITAL_PRIVATE_KEY` is empty it generates a bare base64 key with `ExportRSAPrivateKey()`, but `src/external-api/Util/JwtHandler.cs` now uses `ImportFromPem`, which expects PEM text with headers. Could we either make the stub generate PEM instead, or keep a base64 fallback in `JwtHandler`?

IgiCodes (changes requested):

src/external-api/Util/JwtHandler.cs
StuwiiDev: I've fixed it by getting the stub to use PEM. The JwtHandler is a better version because we are only using PEM for this application

IgiCodes (approved):
LGTM ✅

PR #363, opened 2026-06-22
feat: SUI-1847 - support configurable reconciliation storage jobs

IgiCodes:
## Summary

Adds configurable reconciliation-mode processing for storage jobs so incoming CSV schemas can be mapped at runtime without baking source-specific column names into the image or repository.

Existing matching behaviour remains the default. Deployments can opt into reconciliation mode when derived comparison fields are required in the output.

## Changes

- Added reconciliation input parsing and orchestration for CSV records, including configurable source address and NHS number handling for derived fields.
- Added runtime `StorageProcessJob:ProcessingMode` selection and protected deployment-time configuration support for storage job settings.
- Made storage job runtime configuration required for blob-event-processor deployments so missing mappings fail early.
- Added explicit source address-history parser selection with `CsvMatchData__AddressHistoryFormat`, supporting the existing `TildePipeChronological` format and the new `SemicolonCommaNewestFirst` format.
- Updated full result export to include reconciliation-derived DOB, address, and NHS-number fields when reconciliation mode is enabled.
- Added generic schema coverage tests for all supported mapped source columns, plus parser/client/orchestrator regression coverage.

## Validation

- Focused unit tests covering parser, CSV processor, reconciliation orchestrator, and full-results export paths.
- E2E reconciliation test.
- Manual local tests using local stub APIs.
- `az bicep build --file infra/stacks/blob-event-processor/subscription.bicep`

## Rollout notes

- Existing deployments continue to use matching mode unless reconciliation mode is explicitly configured.
- Source-specific column mappings should be supplied through protected deployment/environment configuration, not committed to the repo.
- Configure `CsvMatchData__AddressHistoryFormat=SemicolonCommaNewestFirst` where source address history is semicolon-delimited, comma-line formatted, and newest-first.
- Reconciliation mode should be tested in the target environment after a new storage job image has been built.

### DIFF
--- a/.github/workflows/gh-blob-event-processor-infra-deploy.yml
+++ b/.github/workflows/gh-blob-event-processor-infra-deploy.yml
@@ -96,6 +96,7 @@ env:
   TARGET_RESOURCE_GROUP_NAME: ${{ inputs.target_resource_group_name }}
   STORAGE_ACCOUNT_MODE: ${{ inputs.storage_account_mode || 'create' }}
   EXISTING_STORAGE_ACCOUNT_NAME: ${{ inputs.existing_storage_account_name }}
+  # Column mappings may identify a specific deployment, so keep them in secrets to avoid leaking them in workflow logs.
   STORAGE_PROCESS_JOB_CONFIGURATION: ${{ secrets.STORAGE_PROCESS_JOB_CONFIGURATION }}
 
 concurrency:
@@ -124,6 +125,15 @@ jobs:
 
           if [ -z "${AZURE_CONTAINER_APP_PE_SUBNET}" ]; then
             echo "::error::AZURE_CONTAINER_APP_PE_SUBNET secret must be configured for blob-event-processor infrastructure deployments."
+            exit 1
+          fi
+
+      - name: Validate storage process job configuration
+        run: |
+          set -euo pipefail
+
+          if [ -z "${STORAGE_PROCESS_JOB_CONFIGURATION}" ]; then
+            echo "::error::STORAGE_PROCESS_JOB_CONFIGURATION secret must be configured for blob-event-processor infrastructure deployments."
             exit 1
           fi
 
--- a/.github/workflows/gh-blob-event-processor-infra-deploy.yml
+++ b/.github/workflows/gh-blob-event-processor-infra-deploy.yml
@@ -96,6 +96,7 @@ env:
   TARGET_RESOURCE_GROUP_NAME: ${{ inputs.target_resource_group_name }}
   STORAGE_ACCOUNT_MODE: ${{ inputs.storage_account_mode || 'create' }}
   EXISTING_STORAGE_ACCOUNT_NAME: ${{ inputs.existing_storage_account_name }}
+  STORAGE_PROCESS_JOB_CONFIGURATION: ${{ secrets.STORAGE_PROCESS_JOB_CONFIGURATION }}
 
 concurrency:
   group: blob-event-processor-${{ inputs.environment_prefix }}-${{ inputs.environment }}
@@ -214,6 +215,7 @@ jobs:
           target_resource_group_name: ${{ env.TARGET_RESOURCE_GROUP_NAME }}
           storage_account_mode: ${{ env.STORAGE_ACCOUNT_MODE }}
           existing_storage_account_name: ${{ env.EXISTING_STORAGE_ACCOUNT_NAME }}
+          storage_process_job_configuration: ${{ env.STORAGE_PROCESS_JOB_CONFIGURATION }}
 
   deploy:
     name: Deploy blob-event-processor infrastructure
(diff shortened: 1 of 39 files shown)


.github/workflows/gh-blob-event-processor-infra-deploy.yml
```diff
@@ -96,6 +96,7 @@ env:
   TARGET_RESOURCE_GROUP_NAME: ${{ inputs.target_resource_group_name }}
   STORAGE_ACCOUNT_MODE: ${{ inputs.storage_account_mode || 'create' }}
   EXISTING_STORAGE_ACCOUNT_NAME: ${{ inputs.existing_storage_account_name }}
+  STORAGE_PROCESS_JOB_CONFIGURATION: ${{ secrets.STORAGE_PROCESS_JOB_CONFIGURATION }}
```
StuwiiDev, line 100: STORAGE_PROCESS_JOB_CONFIGURATION is coming from secrets on this line, but the rest in this file are env.

infra/modules/blob-event-processor/container-app-job.bicep
```diff
@@ -43,6 +43,10 @@ param imageTag string
 @description('The base address of the matching API')
 param matchApiBaseAddress string
 
+@secure()
+@description('Optional runtime configuration values for the storage process job.')
+param storageProcessJobConfiguration object = {}
```
StuwiiDev: non-blocking: I would say this should be mandatory, the service simply won't work without any mapped config? if I have understood this correctly

infra/stacks/blob-event-processor/main.bicep
```diff
@@ -67,6 +67,10 @@ param tagEnvironmentName string = ''
 @description('Optional additional tags to apply to deployed resources.')
 param additionalTags object = {}
 
+@secure()
+@description('Optional runtime configuration values for the storage process job.')
+param storageProcessJobConfiguration object = {}
```
StuwiiDev: non-blocking: same comment as above about needing values

src/SUI.Client/SUI.Client.Core/Application/UseCases/ReconcilePeople/AddressDelimiterParser.cs
```diff
@@ -66,6 +71,48 @@ public static AddressHistory ParseHistory(string? historyString, string? primary
         return new AddressHistory(addresses, primaryAddress);
     }
 
+    private static AddressHistory ParseNewestFirstCommaSeparatedHistory(
+        string historyString,
+        string primaryPostcode
+    )
+    {
+        var addresses = historyString
+            .Split(';', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries)
```
StuwiiDev: just checking this split by semi colon is a requirement going forward on splitting address history?

StuwiiDev (commented):
looks fine, all minor non blocking comments/questions

.github/workflows/gh-blob-event-processor-infra-deploy.yml
IgiCodes: This is intentional. The column mappings may contain information that could identify a specific deployment and we want to avoid identifiable column names from being leaked into workflow logs :slightly_smiling_face: 

I'll add a quick comment or something to explain that to avoid any future confusion as I feel that it's quite important :ok_hand:

infra/modules/blob-event-processor/container-app-job.bicep
IgiCodes: Agreed, will adjust the descriptions and remove default values :ok_hand:

infra/stacks/blob-event-processor/main.bicep
IgiCodes: Agreed, will adjust the descriptions and remove default values :ok_hand:

src/SUI.Client/SUI.Client.Core/Application/UseCases/ReconcilePeople/AddressDelimiterParser.cs
IgiCodes: This came from the example/schema spreadsheet we were provided. The sample Address history value has multiple addresses separated by semicolons and the notes say the most recent address is first. The existing `~`/ `|` format still works, this just adds support for `;` also :slightly_smiling_face: 

I will check for clarity on this though to make sure that's an intentional difference. Good shout :ok_hand:

.github/workflows/gh-blob-event-processor-infra-deploy.yml
IgiCodes: Comment added for clarification 👌

infra/modules/blob-event-processor/container-app-job.bicep
IgiCodes: Made required config now

infra/stacks/blob-event-processor/main.bicep
IgiCodes: Made required config now

src/SUI.Client/SUI.Client.Core/Application/UseCases/ReconcilePeople/AddressDelimiterParser.cs
IgiCodes: Got clarification that the new format is intentional.

Improved the code around this to be config driven and to utilise interface implementations for the different address formats.

StuwiiDev (approved):

PR #364, opened 2026-06-30
sui-1814 - Implement GraphQL Data Extraction from Eclipse

tomhawkin:
## Summary

Added a new job to get data from Eclipse GraphQL endpoint. 

## Changes

- new console app for fetching graphql data
- fake graphQL api for local testing
- unit tests

## Validation

- unit tests
- ran against ecplise sandbox api
-


src/SUI.Client/SUI.Client.GraphQLProcessJob/GraphQLProcessJobOptions.cs
```diff
@@ -0,0 +1,15 @@
+namespace SUI.Client.GraphQLProcessJob;
+
+public sealed class GraphQlProcessJobOptions
+{
+    public const string SectionName = "GraphQLProcessJob";
+
+    public string? Url { get; init; }
+    public string? TenantId { get; init; }
+    public string? ClientId { get; init; }
+    public string? ClientSecret { get; init; }
```
jtaylor100, line 10: Thought to suggest fetching the secret from keyvault during execution, though since the job is short-lived this is probably fine.

Looks like either approach is straightforward to implement in bicep.

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/AzureAdAuthHandler.cs
```diff
@@ -0,0 +1,105 @@
+using System.Net.Http.Headers;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+using Microsoft.Extensions.Options;
+
+namespace SUI.Client.GraphQLProcessJob.Infrastructure;
+
+public class AzureAdAuthHandler : DelegatingHandler
```
jtaylor100: Could [`Azure.Identity.ClientSecretCredential`](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.clientsecretcredential?view=azure-dotnet) do the same work here?

I believe it does token fetching and caching. Would need the `Azure.Identity` nuget package. Is there anything special in here that needs a custom client for the token?

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
```diff
@@ -0,0 +1,69 @@
...
+using SUI.Client.Core.Application.Models;
+
+namespace SUI.Client.GraphQLProcessJob.Infrastructure;
+
+public class GraphQlProcessor(
+    ILogger<GraphQlProcessor> logger,
+    IEclipseClient eclipseClient,
+    IOptions<GraphQlProcessJobOptions> options)
+{
+    public async Task RunAsync(CancellationToken cancellationToken)
```
jtaylor100, line 18: I think the only way to address this sonar error would be to break the code into more methods.

But I think it's actually easier to read altogether like this, so happy to ignore this check failure.

src/SUI.Client/SUI.Client.GraphQLProcessJob/.graphqlrc.json
```diff
@@ -0,0 +1,24 @@
+{
+  "schema": "schema.graphql",
+  "documents": "**/*.graphql",
+  "extensions": {
+    "strawberryShake": {
+      "name": "EclipseClient",
+      "namespace": "Eclipse.GraphQL",
+      "url": "https://olm-daisy-ecm.olmapps.com/integration",
```
jtaylor100, line 8: I think from reading `Program.cs` the requesting Url can be varied at runtime. Does strawberry shake let you remove the URL from this config, in that case? Just don't want this to mislead on where the requesting URL is configured, since this will vary by environment.

src/SUI.Client/SUI.Client.GraphQLProcessJob/graphql.config.yml
```diff
@@ -0,0 +1,6 @@
+schema: schema.graphql
+extensions:
+  endpoints:
+    production:
+      url: https://olm-daisy-ecm.olmapps.com/integration
+documents: '**/*.graphql'
```
jtaylor100: Is this a duplicate of `.graphqlrc.json`? Or maybe the same contents can be put into the 'extensions' section of `.graphqlrc.json`?

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                    if (logger.IsEnabled(LogLevel.Information))
+                    {
+                        logger.LogInformation(
+                            "Person Id: {ID}, Name: {Forename} {Surname}, DOB: {DateOfBirth}, Gender: {Gender}, NHS No: {NHSNumber}, Postcode: {Postcode}",
+                            person.Id, person.Forename, person.Surname, person.DateOfBirth?.Lower, person.Gender,
+                            person.NhsNumber,
+                            person.Addresses.FirstOrDefault(a => a.Id == person.PreferredAddress?.Id)?.Location
+                                ?.Postcode);
+                    }
```
jtaylor100, line 48: I guess around here is where the call to `IMatchPersonRecordOrchestrator` will be. Needless to say, you're planning to remove this log, right?

jtaylor100 (commented):
A few comments, but I think the only one that might have a big change on the `AzureAdAuthHandler`

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
tomhawkin: yeah this is just so that i can test its working whilst it doesn't actually do anything useful

src/SUI.Client/SUI.Client.GraphQLProcessJob/graphql.config.yml
tomhawkin: the .graphqlrc.json is for strawberry shake to generate the client. the graphql.config.yml is for the rider graphql addon to be able to see the api and help you explore it and generate potential queries. we could get rid of that one if you want as its just a convenience.

src/SUI.Client/SUI.Client.GraphQLProcessJob/.graphqlrc.json
tomhawkin: this is used at build time to generate the client. if we need to generate from production then we would want to use the production url here as well really but we don't have it yet

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
tomhawkin: i may refactor it when i do the next ticket

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/AzureAdAuthHandler.cs
tomhawkin: i have used azure identity to simplify the handler. i kept the handler though because it seemed cleaner than setting it all up in program.cs

src/SUI.Client/SUI.Client.GraphQLProcessJob/.graphqlrc.json
jtaylor100: Okay, sounds reasonable. We can have the production URL be a secret value injected in at deployment -- since the deployment will also involve building the container image.

src/SUI.Client/SUI.Client.GraphQLProcessJob/graphql.config.yml
jtaylor100: I really think from reading this that they are the same config file: https://the-guild.dev/graphql/config/docs/user/usage

Have you tried adding something like this to .graphqlrc.json:
```
{
    ...
    "extensions": {
        "endpoints": {
            "production": {
                "url": ...
            }
        },
        ... strawberry shake etc ...
    }
}
```
then removing the yml file, and seeing if the IDE picks it up in the same way?

If that doesn't work, then let's not bother and we can resolve this.

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/AzureAdAuthHandler.cs
jtaylor100: Looks good to me, thanks pal

src/SUI.Client/SUI.Client.GraphQLProcessJob/graphql.config.yml
tomhawkin: i have removed it and everything seems to still work

jtaylor100 (approved):

PR #365, opened 2026-06-30
docs: SUI-1838 - document deployment stacks recommendation

IgiCodes:
## Summary

Documents the SUI-1838 investigation into whether SUI Matcher should adopt Azure Deployment Stacks.

Deployment Stacks provide useful resource ownership and selective teardown capabilities, particularly when deploying into a client-owned resource group. The recommendation is not to adopt them within the current project timelines because environment teardown is infrequent and supporting both SUI-owned and client-owned deployment models would not provide enough value to justify the implementation and operational effort.

## Changes

- Add a design note covering Deployment Stack capabilities, benefits, risks, alternatives, and what-if mitigations.
- Compare SUI-owned resource groups with existing client-owned resource groups and resources.
- Recommend retaining the current Bicep deployment model, with criteria for revisiting a targeted Deployment Stack pilot.
- Link the spike findings from the infrastructure README.

## Validation

- Ran `poetry run mkdocs build --strict`.
- Documentation build completed successfully.
- Documentation-only change; no runtime validation required.

### DIFF
--- a/documentation/docs/design/azure-deployment-stacks-spike.md
+++ b/documentation/docs/design/azure-deployment-stacks-spike.md
@@ -1,0 +1,235 @@
+# SUI-1838: Azure Deployment Stacks Spike
+
+**Date:** `2026-06-26`
+
+**Status:** Findings and recommendation
+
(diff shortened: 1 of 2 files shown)


jtaylor100 (approved):
Thanks for your work and for the new knowledge 👍

PR #367, opened 2026-07-06
SUI-1816: Added Matching

tomhawkin:
## Summary

added code to deal with the GraphQL response and pass it into the existing matching code

## Changes

-GraphQLProcessor now runs matching on all the items returned from GraphQL
-
-

## Validation

- unit tests
- ran code locally
-

### DIFF
--- a/src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
+++ b/src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
@@ -47,7 +47,7 @@ public class GraphQlProcessor(
                         { mappings.Id, person.Id },
                         { mappings.Given, person.Forename ?? "" },
                         { mappings.Family, person.Surname ?? "" },
-                        { mappings.BirthDate, person.DateOfBirth?.Lower?.ToString("dd/MM/yyyy") ?? "" },
+                        { mappings.BirthDate, person.DateOfBirth?.Lower?.ToString(csvMatchDataOptions.Value.DateFormat) ?? "" },
                         {
                             mappings.Postcode,
                             person.Addresses.FirstOrDefault(a => a.Id == person.PreferredAddress?.Id)?.Location
@@ -62,7 +62,7 @@ public class GraphQlProcessor(
 
                     if (!string.IsNullOrEmpty(mappings.Gender))
                     {
-                        personDictionary[mappings.Gender] = person.Gender?.ToString() ?? "";
+                        personDictionary[mappings.Gender] = person.Gender?.ToString().ToLower() ?? "";
                     }
 
                     csvRecords.Add(new CsvRecordDto(personDictionary));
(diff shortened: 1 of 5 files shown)


src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
```diff
@@ -37,18 +42,33 @@ public async Task RunAsync(CancellationToken cancellationToken)
...
+                    var personDictionary = new Dictionary<string, string>
                     {
-                        logger.LogInformation(
-                            "Person Id: {ID}, Name: {Forename} {Surname}, DOB: {DateOfBirth}, Gender: {Gender}, NHS No: {NHSNumber}, Postcode: {Postcode}",
-                            person.Id, person.Forename, person.Surname, person.DateOfBirth?.Lower, person.Gender,
-                            person.NhsNumber,
+                        { mappings.Id, person.Id },
+                        { mappings.Given, person.Forename ?? "" },
+                        { mappings.Family, person.Surname ?? "" },
+                        { mappings.BirthDate, person.DateOfBirth?.Lower?.ToString("dd/MM/yyyy") ?? "" },
```
IgiCodes: **issue (blocking):** Could we use `csvMatchDataOptions.Value.DateFormat` here instead of hard-coding `dd/MM/yyyy`? The record is passed straight into `CsvPersonSpecParser`, which parses DOB using the configured `CsvMatchData:DateFormat`, so any deployment using a different format would send records to matching with `BirthDate = null`.

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
```diff
@@ -37,18 +42,33 @@ public async Task RunAsync(CancellationToken cancellationToken)
...
+                    };
+
+                    if (!string.IsNullOrEmpty(mappings.NhsNumber))
+                    {
+                        personDictionary[mappings.NhsNumber] = person.NhsNumber ?? "";
+                    }
+
+                    if (!string.IsNullOrEmpty(mappings.Gender))
+                    {
+                        personDictionary[mappings.Gender] = person.Gender?.ToString() ?? "";
```
IgiCodes: **issue (blocking):** I think this needs mapping to the lower-case FHIR gender values expected by `PersonSpecification` (`male`, `female`, `unknown`, `other`). `person.Gender?.ToString()` gives the generated enum names like `Male` / `Female`, which the matching API treats as invalid and strips before search.

IgiCodes (changes requested):

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
tomhawkin: added a toLower

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
tomhawkin: changed this over to the config value

IgiCodes (approved):

PR #370, opened 2026-07-14
SUI-1820 - Write Confident matches back to eclipse

tomhawkin:
## Summary

added UpdatePerson mutation to save strong match NHS numbers back to eclipse.
Refactored GraphQLProcessor

## Changes

- split GraphQlProcessor method up to reduce complexity
- added mutation to save back nhs numbers

## Validation

- unit tests
- run against fake api

### DIFF
--- a/src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
+++ b/src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
@@ -1,3 +1,5 @@
+using System.Diagnostics;
+
 using Eclipse.GraphQL;
 
 using Microsoft.Extensions.Logging;
@@ -19,15 +21,18 @@ public class GraphQlProcessor(
     IOptions<GraphQlProcessJobOptions> options,
     IOptions<CsvMatchDataOptions> csvMatchDataOptions)
 {
+    private record PersonMetadata(int ObjectVersion, string? NhsNumber, IReadOnlyList<PersonType> PersonTypes);
+
     public async Task RunAsync(CancellationToken cancellationToken)
     {
+        var timer = Stopwatch.StartNew();
         logger.LogInformation("Running Graph QL Process Job.");
         var mappings = csvMatchDataOptions.Value.ColumnMappings;
 
-        var (csvRecords, personObjectVersions) = await FetchAndCompilePersonRecordsAsync(mappings, cancellationToken);
+        (List<CsvRecordDto> csvRecords, Dictionary<string, PersonMetadata> personMetadata) = await FetchAndCompilePersonRecordsAsync(mappings, cancellationToken);
 
-        logger.LogInformation("Completed compiling GraphQL records. Total records retrieved: {Count}.",
-            csvRecords.Count);
+        logger.LogInformation("Completed compiling GraphQL records. Total records retrieved: {Count}. Elapsed Time: {ElapsedTime}",
+            csvRecords.Count, timer.Elapsed.ToString("g"));
 
         var matchedResults = await matchPersonRecordOrchestrator.ProcessAsync(
             csvRecords,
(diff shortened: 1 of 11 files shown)


src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
```diff
@@ -21,83 +22,144 @@ public class GraphQlProcessor(
...
+
+        await SaveMatchedNhsNumbersAsync(matchedResults, personObjectVersions, mappings, cancellationToken);
+    }
+
+    private async Task<(List<CsvRecordDto> CsvRecords, Dictionary<string, int> PersonObjectVersions)> FetchAndCompilePersonRecordsAsync(
+        CsvMatchDataOptions.Headers mappings,
+        CancellationToken cancellationToken)
+    {
         int pageNumber = 1;
         const int pageSize = 10;
```
jtaylor100: Didn't ask before but, any particular reason for a page size of 10? Or is does the page size not matter too much, just as long as its not the full result in one response?

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
```diff
@@ -21,83 +22,144 @@ public class GraphQlProcessor(
...
+            logger.LogInformation("Saving matched NHS number {NhsNumber} for Person {PersonId} with ObjectVersion {ObjectVersion}.",
+                matchedNhsNumber, personId, objectVersion);
+
+            try
+            {
+                var updateInput = new UpdatePerson
+                {
+                    Id = personId,
+                    NhsNumber = matchedNhsNumber,
+                    ObjectVersion = objectVersion
```
jtaylor100: Any ideas how these object versions work? I'd imagine they are version control on the records, so a new one should be generated each time an update has happened.

However I guess you've passed it back here so the mutation works. Does it the object version change after you do a mutation in the sandbox, passing in the current object version here?

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
tomhawkin: page size can be anything we want. currently it is 10 so that i can easily confirm it is working as expected but we should work out what is a sensible number would be. maybe more like 100.  we should also discuss what the total records are in production and look at how long the job is likely to take. if there are many records we may want to move to a more parallel approach but then this would fire more requests at FHIR at once and we will get rate limited there so there are tradeoffs

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
tomhawkin: im not 100% sure this is working correctly but what it currently does is get the object version from the record and pass that into the mutation. from what i could see it uses the object version as a way of making sure the caller of the mutation had the latest version when they called and if between getting the object and mutating it someone else had mutated it then the object version would be wrong and the mutation would fail

jtaylor100: So the way it's implemented, what happens when a mutation fails?

jtaylor100: Ah wait, sorry, saw the try...catch. So it will skip failed mutations. Sounds good, they can be picked up on the next run if they fail due to a mid-air collision.

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
```diff
@@ -19,15 +21,18 @@ public class GraphQlProcessor(
     IOptions<GraphQlProcessJobOptions> options,
     IOptions<CsvMatchDataOptions> csvMatchDataOptions)
 {
+    private record PersonMetadata(int ObjectVersion, string? NhsNumber, IReadOnlyList<PersonType> PersonTypes);
```
jtaylor100: Isn't this data already available in scope within `ProcessedMatchedRecord<CsvRecordDto>.OriginalData["nhsNumber"]` or similar? On the matched results being looped through in `SaveMatchedNhsNumbersAsync`.

Did you need a separate datatype for a different reason?

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
```diff
@@ -127,28 +145,37 @@ private async Task SaveMatchedNhsNumbersAsync(
...
+            {
+                logger.LogWarning("Could not find metadata for Person {PersonId}. Skipping NHS number update.", personId);
+                continue;
+            }
+
+            if (!string.IsNullOrEmpty(metadata.NhsNumber))
             {
-                logger.LogWarning("Could not find ObjectVersion for Person {PersonId}. Skipping NHS number update.", personId);
+                logger.LogInformation("Person {PersonId} already has NHS number {ExistingNhsNumber}. Skipping update.",
+                    personId, metadata.NhsNumber);
```
jtaylor100: Please no logging of the NHS number!

src/SUI.Client/SUI.Client.GraphQLProcessJob/PersonByCriteria.graphql
```diff
@@ -23,6 +23,7 @@ query PersonByCriteria($maxAge: Int, $paging: RequestCursorInput) {
                 }
                 id
                 objectVersion
+                personTypes
```
jtaylor100, line 26: What did you need the person type for?

tomhawkin: when i tried mutating records against the eclipse api i was getting validation failures stating that the personTypes was empty and it must contain a value so i just pulled it in the query and pass it back in on the mutation.

jtaylor100 (approved):

PR #373, opened 2026-07-17
SUI-1823 - Configure ACA Scheduled Job

tomhawkin:
## Summary

added api batch processor deploy workflow. added graphql container job

## Changes

- added container job to api-batch-processor stack
- added deploy workflow for api-batch-processor stack

## Validation

- bicep build
- run deploy after merge
-

### DIFF
--- a/.github/actions/stack-infra-run/action.yml
+++ b/.github/actions/stack-infra-run/action.yml
@@ -108,6 +108,22 @@ inputs:
     description: "Optional JSON object containing protected storage process job runtime configuration"
     required: false
     default: ""
+  graphql_process_job_image_tag:
+    description: "Optional GraphQL process job image tag"
+    required: false
+    default: ""
+  deployment_mode:
+    description: "Optional deployment mode: automatic or manual"
+    required: false
+    default: ""
+  cron_expression:
+    description: "Optional cron expression for scheduled runs"
+    required: false
+    default: ""
+  graphql_process_job_configuration:
+    description: "Optional JSON object containing protected GraphQL process job runtime configuration"
+    required: false
+    default: ""
 
 
 runs:
@@ -132,6 +148,7 @@ runs:
       env:
         ADDITIONAL_TAGS: ${{ inputs.additional_tags }}
         STORAGE_PROCESS_JOB_CONFIGURATION: ${{ inputs.storage_process_job_configuration }}
+        GRAPHQL_PROCESS_JOB_CONFIGURATION: ${{ inputs.graphql_process_job_configuration }}
       run: |
         set -euo pipefail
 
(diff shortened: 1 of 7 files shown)


src/SUI.Client/SUI.Client.GraphQLProcessJob/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM mcr.microsoft.com/dotnet/sdk:10.0 AS build
+WORKDIR /src
+
+COPY global.json Directory.Build.props Directory.Packages.props NuGet.config ./
+COPY src/SUI.Client/SUI.Client.Core/SUI.Client.Core.csproj src/SUI.Client/SUI.Client.Core/
```
IgiCodes, line 6: **issue (blocking):** This Dockerfile needs to copy `src/Shared/Shared.csproj` and `src/Shared/` as well. `SUI.Client.Core.csproj` has a project reference to `../../Shared/Shared.csproj`, so the image build currently fails.

infra/stacks/api-batch-processor/main.bicep
```diff
@@ -1,41 +1,164 @@
...
 var stackNameSuffix = 'abp'
-var noHyphensEnvironmentPrefix = replace(environmentPrefix, '-', '')
-var storageAccountName = toLower('${take(noHyphensEnvironmentPrefix, 8)}${take(lowercaseEnvironmentName, 8)}${stackNameSuffix}${take(uniqueString(resourceGroup().id, noHyphensEnvironmentPrefix, lowercaseEnvironmentName), 5)}')
-var tags = {
+var isProductionEnvironment = lowercaseEnvironmentName == 'prod' || lowercaseEnvironmentName == 'production'
+var allowedNhsFqdns = isProductionEnvironment ? [
+  'api.service.nhs.uk'
+] : [
+  'int.api.service.nhs.uk'
+]
```
IgiCodes: **issue (blocking):** The GraphQL job gets its URL from `GraphQLProcessJob__Url`, but this stack routes CAE egress through Azure Firewall and only allows `api.service.nhs.uk` / `int.api.service.nhs.uk` for workload API traffic. Azure Firewall blocks outbound FQDNs that don’t match an allow rule, so if the GraphQL API lives anywhere else this job won’t be able to reach it. Should we add a configurable allowed GraphQL FQDN list here too?

.github/workflows/gh-api-batch-processor-infra-deploy.yml
```diff
@@ -0,0 +1,292 @@
...
+          tag_environment_name: ${{ env.AZURE_TAG_ENVIRONMENT_NAME }}
+          additional_tags: ${{ env.AZURE_ADDITIONAL_TAGS }}
+          azure_include_role_assignments: ${{ env.AZURE_INCLUDE_ROLE_ASSIGNMENTS }}
+          stack_resource_group: ${{ env.STACK_RESOURCE_GROUP }}
+          deployment_name: ${{ env.DEPLOYMENT_NAME }}
+          turn_on_alerts: ${{ env.AZURE_TURN_ON_ALERTS }}
+          graphql_process_job_image_tag: ${{ env.GRAPHQL_PROCESS_JOB_IMAGE_TAG }}
+          deployment_mode: ${{ env.DEPLOYMENT_MODE }}
+          cron_expression: ${{ env.CRON_EXPRESSION }}
+          graphql_process_job_configuration: ${{ env.GRAPHQL_PROCESS_JOB_CONFIGURATION }}
```
IgiCodes, line 212: **issue (blocking):** The workflow accepts `resource_group_mode` and `target_resource_group_name`, and `prepare` uses them to derive `STACK_RESOURCE_GROUP`, but this action never forwards them. Because `subscription.bicep` defaults `resourceGroupMode` to `create`, running the workflow with `existing` still deploys the template in create mode. The deploy step has the same issue.

.github/workflows/gh-api-batch-processor-infra-deploy.yml
```diff
@@ -0,0 +1,292 @@
...
```
IgiCodes, line 258: **issue (blocking):** Same issue as above: the deploy action call also needs to forward `resource_group_mode` and `target_resource_group_name`, otherwise the `existing` resource-group option is ignored for actual deployments too.

IgiCodes (changes requested):
Few issues but overall looking good

.github/actions/stack-infra-run/action.yml
```diff
@@ -226,6 +230,11 @@ runs:
           PARAMETERS+=("graphqlProcessJobConfiguration=${GRAPHQL_PROCESS_JOB_CONFIGURATION_JSON}")
         fi
 
+        if [ -n "${{ inputs.allowed_graphql_fqdns }}" ] && [ "${{ inputs.allowed_graphql_fqdns }}" != "[]" ]; then
+          ALLOWED_GRAPHQL_FQDNS_JSON="$(jq -nce --argjson fqdns "${{ inputs.allowed_graphql_fqdns }}" 'if ($fqdns | type) == "array" then $fqdns else error("allowed_graphql_fqdns must be a JSON array") end')"
+          PARAMETERS+=("allowedGraphQLFqdns=${ALLOWED_GRAPHQL_FQDNS_JSON}")
+        fi
```
IgiCodes: **issue (if-minor):** This new JSON input is interpolated directly into the shell script. If the workflow passes a normal JSON array like `["graphql.example.com"]`, the embedded quotes are consumed by shell parsing before `jq --argjson` sees them, so the parse fails. The other JSON inputs avoid this by first putting the value in an env var and then passing `"${VAR}"` to jq. Could we do the same for `allowed_graphql_fqdns`?

IgiCodes (approved):

PR #375, opened 2026-07-22
SUI-1823 - added matching and external api to api-batch-processor stack

tomhawkin:
## Summary

added matching and external api to api-batch-processor stack

## Changes

- added matching and external api to api-batch-processor stack

## Validation

- bicep build

### DIFF
--- a/.github/workflows/gh-api-batch-processor-graphql-job-image.yml
+++ b/.github/workflows/gh-api-batch-processor-graphql-job-image.yml
@@ -119,7 +119,6 @@ jobs:
             --image "${IMAGE_REPOSITORY}:${{ steps.context.outputs.image_tag }}" \
             --image "${IMAGE_REPOSITORY}:latest" \
             --file src/SUI.Client/SUI.Client.GraphQLProcessJob/Dockerfile \
-            --build-arg MATCH_API_BASE_ADDRESS="${{ steps.context.outputs.match_api_base_address }}" \
             .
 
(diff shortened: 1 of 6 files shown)


.github/workflows/gh-api-batch-processor-graphql-job-image.yml
```diff
@@ -0,0 +1,140 @@
...
+      - name: Build GraphQL process job image in ACR
+        run: |
+          set -euo pipefail
+
+          az acr build \
+            --registry "${{ steps.context.outputs.acr_name }}" \
+            --image "${IMAGE_REPOSITORY}:${{ steps.context.outputs.image_tag }}" \
+            --image "${IMAGE_REPOSITORY}:latest" \
+            --file src/SUI.Client/SUI.Client.GraphQLProcessJob/Dockerfile \
+            --build-arg MATCH_API_BASE_ADDRESS="${{ steps.context.outputs.match_api_base_address }}" \
```
IgiCodes: **nitpick (non-blocking):** looks like we can drop `MATCH_API_BASE_ADDRESS` from the GraphQL image build. The Dockerfile doesn’t use it, and the job already gets the URL from the Bicep env var at runtime, which seems like the better place for it anyway.

IgiCodes (dismissed):
Looks good! One very minor comment but it's not affecting anything, more just cleanup :)

IgiCodes (approved):

PR #380, opened 2026-07-30
SUI-1911 - added work list definition filter

tomhawkin:
## Summary

added a filter that if the setting is set will only allow records with the specific work list definition id to be added for matching

## Changes

- added work item definition to fake graph ql api
- added filter to graphql processor
- added more unit tests

## Validation

- unit tests
- run against fake graph ql api
- run against sandbox graphql api

### DIFF
--- a/src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
+++ b/src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
@@ -72,16 +72,14 @@ public class GraphQlProcessor(
 
             foreach (var result in resultsList)
             {
-                if (result is IPersonByCriteria_PersonByCriteria_Results_Person person)
+                if (result is not IPersonByCriteria_PersonByCriteria_Results_Person person)
                 {
-                    bool shouldProcess = string.IsNullOrEmpty(options.Value.KnownSafeguardingConcernId) ||
-                        (person.WorklistInstances != null &&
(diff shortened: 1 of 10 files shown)


src/SUI.Client/SUI.Client.GraphQLProcessJob/appsettings.Development.json
```diff
@@ -1,6 +1,7 @@
 {
   "GraphQlProcessJob": {
     "Url": "http://localhost:5050/integration/",
-    "MatchApiBaseAddress": "http://localhost:5000/matching/"
+    "MatchApiBaseAddress": "http://localhost:5000/matching/",
+    "KnownSafeguardingConcernId": "51D2D445-AFF8-40D8-913F-148479653AA6"
```
IgiCodes: **nitpick (non-blocking):** could we rename this to `KnownSafeguardingConcernWorklistDefinitionId`?
`KnownSafeguardingConcernId` reads a bit like the ID of a safeguarding concern record, but the code is matching against `worklistDefinition.id` so it just makes it clearer exactly what the purpose of this config value is.

src/SUI.Client/SUI.Client.GraphQLProcessJob/Infrastructure/GraphQLProcessor.cs
IgiCodes, line 53: **issue (if-minor):** SonarCloud is flagging this method for Cognitive Complexity. Could we pull the new worklist filtering condition into a small helper method or something, so the loop stays focused on fetching/iterating records? Looks like might be a quick readability win if easy enough 🙂

IgiCodes (commented):
Looks good, just a few tidy up bits if it's not too much extra effort 👌

IgiCodes (approved):

